PR #8, opened 2018-04-20
Support GitHub token for pushing gh-pages

jirikuncar:
Provide an alternative authentication method for pushing to gh-pages.

### DIFF
--- a/chartpress.py
+++ b/chartpress.py
@@ -15,9 +15,24 @@ from ruamel.yaml import YAML
 
 __version__ = '0.2.0.dev'
 
+# name of the environment variable with GitHub token
+GITHUB_TOKEN_KEY = 'GITHUB_TOKEN'
+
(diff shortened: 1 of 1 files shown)


jirikuncar:
@Carreau thanks! I have fixed a typo in docs.

chartpress.py
```diff
@@ -15,11 +15,26 @@
 
 __version__ = '0.2.0.dev'
 
+# name of the environment variable with GitHub token
+GITHUB_TOKEN_KEY = 'GITHUB_TOKEN'
```
minrk: You've defined this variable, but haven't used it below. Since it's used only once, I think we can drop the variable.

jirikuncar: Thanks!

jirikuncar:
@minrk is there anything else I should fix?

minrk:
Thanks!

PR #14, opened 2018-05-31
Added option to change docker image prefix

leafty:
This change adds the option to overwrite the `imagePrefix` option from `chartpress.yaml`.

This allows to fork, modify and test a project without having to edit `chartpress.yaml`, as you can then publish container images to a different Docker registry and/or repository.


minrk:
Thanks! Merged as `--image-prefix` to be consistent with the yaml value.

PR #17, opened 2018-08-10
Fix build_values mapping check

leafty:
The yaml parser creates objects which do not have `dict` as a parent class, but `collections.abc.MutableMapping`.

See source for `ruamel.yaml.comments.CommentedMap`: https://bitbucket.org/ruamel/yaml/src/c177aa5d95fd92911eba4b40d60ac20b4e940105/comments.py?at=default&fileviewer=file-view-default#comments.py-595

### DIFF
--- a/chartpress.py
+++ b/chartpress.py
@@ -7,8 +7,9 @@ This is used as part of the JupyterHub and Binder projects.
 
 import argparse
 import os
-import subprocess
 import shutil
+import subprocess
+from collections.abc import MutableMapping
 from tempfile import TemporaryDirectory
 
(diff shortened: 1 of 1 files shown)


consideRatio:
@betatim @leafty #18 :D We almost have the exact same PRs

- Should we use MutableMapping or Mapping? [MutableMapping inherits from Mapping](https://docs.python.org/3/library/collections.abc.html#collections-abstract-base-classes).

I figure we should use MutableMapping, as we are modifying the object we check for being either Mapping or MutableMapping, so if it isn't mutable we fail later anyhow!

I'm closing my PR.

PR #20, opened 2018-08-15
Coerce ruamel.yaml seq to be python list

### LINKED ISSUE #19: TypeError: can only concatenate list (not "CommentedSeq") to list
It seems that updates from `ruamel.yaml` continue to affect chartpress.

Related: https://github.com/jupyterhub/chartpress/pull/17

Both issues come from the changes in `ruamel.yaml` changing the hierarchy of yaml objects.

leafty:
Fix #19


betatim:
Looks good to me. Let's merge this now, and discuss pinning in a new issue? I'd be +1 for pinning.

PR #28, opened 2018-12-13
provide a reset option to reset image tags and chart version

rokroskar:
Provide a `--sanitize` option to reset the image tags. This is useful to avoid committing unwanted changes to chart image tags.

### DIFF
--- a/chartpress.py
+++ b/chartpress.py
@@ -191,7 +191,7 @@ def image_needs_building(image):
     return image_needs_pushing(image)
 
 
-def build_images(prefix, images, tag=None, commit_range=None, push=False, chart_version=None):
+def build_images(prefix, images, tag=None, commit_range=None, push=False, chart_version=None, sanitize=False):
     """Build a collection of docker images
 
     Args:
@@ -227,10 +227,20 @@ def build_images(prefix, images, tag=None, commit_range=None, push=False, chart_
         image_name = prefix + name
         image_spec = '{}:{}'.format(image_name, image_tag)
 
+        if sanitize:
+            image_tag = 'latest'
+
         value_modifications[options['valuesPath']] = {
             'repository': image_name,
             'tag': SingleQuotedScalarString(image_tag),
         }
+
+        if sanitize:
+            continue
+
+        if tag is None and commit_range and not path_touched(*paths, commit_range=commit_range):
+            print(f"Skipping {name}, not touched in {commit_range}")
+            continue
 
         template_namespace = {
             'LAST_COMMIT': last_commit,
@@ -289,7 +299,7 @@ def build_values(name, values_mods):
         yaml.dump(values, f)
 
 
-def build_chart(name, version=None, paths=None):
+def build_chart(name, version=None, paths=None, sanitize=False):
     """Update chart with specified version or last-modified commit in path(s)"""
     chart_file = os.path.join(name, 'Chart.yaml')
     with open(chart_file) as f:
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -44,9 +44,9 @@ def run(self):
     license="BSD",
     platforms="Linux, Mac OS X",
     keywords=['helm', 'kubernetes'],
-    python_requires=">=3.4",
+    python_requires=">=3.6",
     install_requires=[
-        'ruamel.yaml>=0.15',
+        'ruamel.yaml<=0.15.54',
```
minrk: dependencies shouldn't be pinned down

chartpress.py
```diff
@@ -289,6 +299,8 @@ def main():
         help='extra message to add to the commit message when publishing charts')
     argparser.add_argument('--image-prefix', default=None,
         help='override image prefix with this value')
+    argparser.add_argument('--sanitize', action='store_true',
```
minrk: Sanitize is a surprising name, and it isn't clear what "clean up image tags" means. Is this resetting to the original state (i.e. the same as `git checkout -- chart`)? If so, I might call it `--reset`, and give a more detailed explanation of what it does.

setup.py
rokroskar: this might be a relic from a few months back -- updates to the `ruamel` library tend to cause unpredictable behavior at times

chartpress.py
rokroskar: yes, `--reset` may be a better name. It's like a `git checkout -- chart` except that in that case it's annoying to have to juggle commits. 

The idea here is really just to undo the image tag and chart version changes of `chartpress` to keep the diff cleaner. I've found it reasonable to keep image tags at `latest` in the codebase, and only change them in published charts. However, that might not work for everyone so maybe specifying a different "reset" tag/version would be the way to go. However, in that case maybe all that is needed is to have a `--skip-build` option with `--tag`.

minrk: putting the resetTag in chartpress.yaml would be a way to allow charts to pick what they use (we use an explicitly invalid `set-by-chartpress`, since installing from git without rendering doesn't make sense for our charts)

rokroskar: +1 on the `set-by-chartpress`. I will implement the `resetTag` option and rename `--sanitize` to `--reset`.

setup.py
minrk: Any time that happens would be a bug to be fixed. New releases of ruamel.yaml shouldn't force new releases of chartpress. If pin-down is desired, it should go in a given project's CI requirements.txt, never a library's strictly enforced dependencies.

rokroskar:
@minrk apologies for the long delay with this PR - I've addressed the comments, please re-review.

fgeorgatos:
@minrk : could we revive this PR? any pending comments?

chartpress.py
```diff
@@ -232,6 +234,13 @@ def build_images(prefix, images, tag=None, commit_range=None, push=False, chart_
             'tag': SingleQuotedScalarString(image_tag),
         }
 
+        if skip_build:
+            continue
+
+        if tag is None and commit_range and not path_touched(*paths, commit_range=commit_range):
```
minrk: This conditional should be removed, I think. This logic is below. Perhaps added by a sync with master?

rokroskar: ah yes, most likely a rebase error, thanks

minrk:
Thanks!

PR #29, opened 2018-12-19
include chart version as prefix in image tags

### LINKED ISSUE #26: Support semantic version for tags
I raised an issue against the zero-to-jupyterhub-k8s project jupyterhub/zero-to-jupyterhub-k8s#1038 but if I've understood the build process correctly the change is probably across both this project (to support version tags) and the zero-to-jupyterhub-k8s project (to modify the CI scripts).

minrk:
so we get `0.8-sha` instead of just `sha`. Makes it a little easier to see which builds are for which chart version

closes #26

### DIFF
--- a/chartpress.py
+++ b/chartpress.py
@@ -117,7 +117,7 @@ def build_image(image_path, image_name, build_args=None, dockerfile_path=None):
     subprocess.check_call(cmd)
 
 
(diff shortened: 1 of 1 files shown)


yuvipanda:
Thanks, @minrk

PR #33, opened 2019-02-19
use git describe for better prerelease chart versions

minrk:
- preserves prerelease category (alpha, beta, etc.) so that each commit after starting the beta cycle will have the beta tag.
- use '0' if no prerelease category is set
- chart version has the form `x.y.z-{alpha|beta|rc|0}.n.sha` where `n` is commits since tag,
  sha is current commit

see https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1154

This does not affect the image tags, which are still of the form `x.y.z-sha`, since sorting is less important there as long as the base version is present.

### DIFF
--- a/chartpress.py
+++ b/chartpress.py
@@ -299,7 +299,19 @@ def build_chart(name, version=None, paths=None):
         if paths is None:
             paths = ['.']
         commit = last_modified_commit(*paths)
-        version = chart['version'].split('-')[0] + '-' + commit
+        # parse prerelease version, preserving only the first part
+        # which will be the prerelease category (alpha, beta, etc.)
+        # if nothing is found, use `0`, to sort before alpha
+        base_version, *rest = chart['version'].split('-', 1)
+        prerelease = 0
+        if rest:
+            prerelease = rest[0].split('.')[0]
+            if not prerelease:
+                prerelease = '0'
+
+        describe = check_output(['git', 'describe', '--tags', '--long', commit]).decode('utf8').strip()
+        _tag, n_commits, sha = describe.rsplit('-', 2)
+        version = "{base_version}-{prerelease}.{n_commits}.{sha}".format(**locals())
 
     chart['version'] = version
 


consideRatio:
The helm chart alpha/beta releases has the format `x.y.z-{alpha|beta|rc}.{n}` (ref: https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1054#issuecomment-461483349).

I'm mentally simulating what happens if we have this system, and have the following pushed chart versions in chronological order.

```
0.8.0
0.8.0-0.4.asdf123
0.8.0-0.10.sdfg234
0.9.0-beta.1
0.9.0-beta.1.dfgh345
0.9.0-beta.5.fghj456
0.9.0-beta.2
0.9.0-beta.1.ghjk567
0.9.0-beta.3
0.9.0
```

To avoid entries like `0.9.0-beta.1.ghjk567` that are actually  made 1 commit after `0.9.0-beta.2`, I suggest to preserve not only `beta` but `beta.n`. We would then instead get...

```
0.8.0
0.8.0-0.4.asdf123
0.8.0-0.10.sdfg234
0.9.0-beta.1
0.9.0-beta.1.1.dfgh345
0.9.0-beta.1.5.fghj456
0.9.0-beta.2
0.9.0-beta.2.1.ghjk567
0.9.0-beta.3
0.9.0
```

consideRatio:
### Final (?) suggestion

Below are [valid SemVer 2 version strings](https://semver.org/#backusnaur-form-grammar-for-valid-semver-versions), which is what [Helm chart's requires](https://helm.sh/docs/charts/#the-chart-yaml-file).

I suggest we find the latest tag*, and construct `f"{latest_tag}+{n_commits:03d}.{sha}"`.

The latest tag is not as trivial as it may seem though ^^. [This stackoverflow answer](https://stackoverflow.com/a/27154277) was essential reading for me. To get the latest commit pointing to a commit on the current branch `git describe --abbrev=0` can be used. For some reason though, in z2jh, 0.8.2 or 0.8.1 does not point to a commit in the master branch. Also one may consider, should we use the latest tagged tag, or latest tagged tag found in the current branch, or latest tagged commit in the current branch? We should use the latest tag found in the current branch, and that is given by `git describe --abbrev=0`.

```
0.8.0
0.8.0+004.asdf123
0.8.0+010.sdfg234
0.9.0
0.9.0-beta.1
0.9.0-beta.1+001.dfgh345
0.9.0-beta.1+005.fghj456
0.9.0-beta.2
0.9.0-beta.2+001.ghjk567
0.9.0-beta.3
```

minrk:
closed by #52

PR #34, opened 2019-05-29
avoid mangling of quotes in rendered charts (#1)

rokroskar:
Some yaml values might be converted to e.g. boolean - for example, an unquoted "off" can get rendered as "false". This has made our nginx-ingress break in a way that was quite difficult to debug and can be avoided with the `preserve_quotes` flag in this PR.


yuvipanda:
Thank you for the PR! I think this makes good sense.

PR #35, opened 2019-06-06
Add optional tag prefix

manics:
Adds an optional prefix to the tag. Example use-case: include the application version in the tag:
- `repository/imagename:app.version-chart.version-commit`


manics:
See https://github.com/jupyterhub/chartpress/issues/37 instead

PR #46, opened 2019-10-12
Refresh .travis.yml

consideRatio:
I don't know for sure how this failed silently, but it didn't actually do what it should before on the z2jh repo.

It contains two commits from #45 about the --version flag.


yuvipanda:
I dig this too

PR #47, opened 2019-10-12
Add RELEASE.md

consideRatio:
This would help someone less experienced, like I have been in this realm, to make a release. It also allows my idea of how to make a release to be revised based on better knowledge in case I have got it wrong.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -7,11 +7,11 @@ make a release on PyPI.
 For you to follow along according to these instructions, you need:
 - To be a maintainer of the [PyPI chartpress
   project](https://pypi.org/project/chartpress/).
-- To have push rights to the chartpress GitHub repository.
+- To have push rights to the [chartpress GitHub repository](https://github.com/jupyterhub/chartpress).
 
 ## Technical steps to make a release
 
-1. Checkout master and make sure your up to date.
+1. Checkout master and make sure it is up to date.
 
    ```
    git checkout master
@@ -48,5 +48,5 @@ For you to follow along according to these instructions, you need:
 1. Push your two commits and tag to master.
 
    ```
-   git push --tags <upstream>
+   git push --tags <upstream> master
    ```
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -45,8 +45,9 @@ For you to follow along according to these instructions, you need:
    appropriately with a `.dev` appendix and make a commit with the message `back
    to dev`.
 
-1. Push your two commits and tag to master.
+1. Push your two commits to master along with the annotated tags referencing
+   commits on master.
 
    ```
-   git push --tags <upstream> master
+   git push --follow-tags <upstream> master
    ```
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,52 @@
+# How to make a release
+
+`chartpress` is a package [available on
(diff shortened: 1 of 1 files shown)


RELEASE.md
```diff
@@ -0,0 +1,52 @@
+# How to make a release
+
+`chartpress` is a package [available on
+PyPI](https://pypi.org/project/chartpress/). These are instructions on how to
+make a release on PyPI.
+
+For you to follow along according to these instructions, you need:
+- To be a maintainer of the [PyPI chartpress
+  project](https://pypi.org/project/chartpress/).
+- To have push rights to the chartpress GitHub repository.
```
manics: In case someone is reading this README in a fork:
```suggestion
- To have push rights to the [chartpress GitHub repository](https://github.com/jupyterhub/chartpress).
```

RELEASE.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Technical steps to make a release
+
+1. Checkout master and make sure your up to date.
```
manics:
```suggestion
1. Checkout master and make sure it is up to date.
```

RELEASE.md
```diff
@@ -0,0 +1,52 @@
...
+   ```
+
+1. Reset the `__version__` variable in [chartpress.py](chartpress.py)
+   appropriately with a `.dev` appendix and make a commit with the message `back
+   to dev`.
+
+1. Push your two commits and tag to master.
+
+   ```
+   git push --tags <upstream>
```
manics: Is this definitely correct?
```
$ git push --help
...
       --tags
           All refs under refs/tags are pushed, in addition to refspecs
           explicitly listed on the command line.
...
```
suggests that won't push master, though it might just be the documentation is unclear.

yuvipanda: I think it'll push them to upstream repo, rather than the master branch there.

To do both, git push --tags upstream master

yuvipanda:
```suggestion
   git push --tags <upstream> master
```

yuvipanda: Actually I was wrong.

git push --tags upstream will actually push to master but also to every other matching branch! See https://makandracards.com/makandra/8039-git-how-to-configure-git-to-push-only-your-current-branch. This might push a bunch of branches you didn't intend.

I woul still recommend explicitly setting name of branch to push

consideRatio: Nice catch!

consideRatio: Hmmm after reading this, I realize that created tags only relate to commits and not branches:

https://stackoverflow.com/questions/14613540/do-git-tags-only-apply-to-the-current-branch/27154277#27154277

And I believe that `git push --tags <remote> <branch>` would push the branch along with tags. I googled about pushing a single tag and found [this stackoverflow question with two important answers](https://stackoverflow.com/questions/5195859/how-do-you-push-a-tag-to-a-remote-repository-using-git).

To push a single tag:

```
git push <remote> <tag>
# or the same done more explicitly
git push <remote> tag <tag>
```

To push all annotated tags that are referencing a commit on branch

```
git push --follow-tags
```

consideRatio:
@manics / @yuvipanda - looks good to you?

yuvipanda:
Thanks, @consideRatio! TIL about --follow-tags too!

PR #52, opened 2019-10-16
Chart and image versioning, and Chart.yaml's --reset interaction

### LINKED ISSUE #44: PR Discussion: chartpress --reset on the Chart.yaml
I want a run of `chartpress --reset` to make someone that has developed in z2jh/binderhub locally be able to run `git add .` and then `git commit` instead of ending up to first clean up the changes of using `chartpress` to develop locally just because chartpress has modified Chart.yaml and values.yaml.

When we run `chartpress --reset`, we modify `Chart.yaml` and `values.yaml`. The configuration option `resetTag` that defaults to `set-by-chartpress` is only used within `values.yaml` while in `Chart.yaml` another logic is used. The logic [is simple](https://github.com/jupyterhub/chartpress/blob/9ded86b3aa9f577d90ae14fe97c8b75c8cbc2b87/chartpress.py#L309-L310):

```python
        if reset:
            version = chart['version'].split('-')[0]
```

In other words, it will take something like 0.9-asdf123 and make it 0.9. We currently have `0.9-dev`. I think we should simply write `0.9` or specifically 0.9.0. Or, make chartpress append `-dev` like we do in z2jh (but [not in binderhub's Helm chart's Chart.yaml](https://github.com/jupyterhub/binderhub/blob/a168d069772012c52f9ac7056ec22d779927ae69/helm-chart/binderhub/Chart.yaml#L1-L4)).

It could be nice to reset to `x.y.z-dev` to make it clear that x.y.z isn't released yet.

consideRatio:
This is a continuation of @minrk's PR #33 where I implement my suggestion in https://github.com/jupyterhub/chartpress/pull/33#issuecomment-542196738.

1. I made the Chart version set itself prefixed with the latest tag found in the branch commit history instead of using a very specific manipulation of the value in Chart.yaml.
1. I made the Chart version set itself suffixed with `+<n_commits_since_tag>.<last_chart_modifying_commit>`, which results in a valid formatted SemVer 2 version.
1. I made the image tags follow the same pattern as the Chart version, but relating to the latest image commit instead of the latest chart commit. There is a difference, and that is that `+` is replaced with `_` to comply with the allowed characters of docker tags.
1. I made the Chart.yaml's `--reset` interaction be to reset the version field to something fixed. This closes #44 using the idea in https://github.com/jupyterhub/chartpress/issues/44#issuecomment-541005642. This change required the first point.

- I also fixed a failure to fail in our TravisCI test caused by multiline test commands that failed or succeeded only based on the last command, as compared to failing on any failure among the multiline commands.

---

## Example diff after chartpress is run on Z2JH

image

Note how it claims that the latest tag is 0.8.0, and that is actually true, it is the latest tagged commit in the git history of z2jh's master. For some reason, 0.8.1 and 0.8.2 tags commit not part of the master commit history. I consider this something that should be fixed in z2jh rather than adjust this PRs chartpress logic which I think makes sense.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,9 +24,8 @@ script:
   - |-
     git clone https://github.com/jupyterhub/zero-to-jupyterhub-k8s &&
     cd zero-to-jupyterhub-k8s &&
-    chartpress &&
-    git --no-pager diff
+    chartpress
+  - git --no-pager diff
   
-  - |-
-    chartpress --reset &&
-    git --no-pager diff
+  - chartpress --reset
+  - git --no-pager diff
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,21 +9,24 @@ script:
   - chartpress --version
   - chartpress --help
   - pyflakes .
-  - |
-    # This is a workaround to an issue caused by the existence of a docker registry
-    # mirror in our CI environment. Without this fix that removes the mirror,
-    # chartpress fails to realize the existence of already built images and rebuilds
-    # them.
-    #
-    # ref: https://github.com/moby/moby/issues/39120
-    echo '{"mtu": 1460}' | sudo dd of=/etc/docker/daemon.json
+
+  # This is a workaround to an issue caused by the existence of a docker registry
+  # mirror in our CI environment. Without this fix that removes the mirror,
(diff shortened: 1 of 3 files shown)


minrk:
Awesome!

PR #53, opened 2019-10-17
Don't append +build on tagged commits

consideRatio:
If you would develop from a tagged commit locally, you would be forced
to rebuild without this change, as the generated tag wouldn't be found
either locally or remotely.

Consider for example `0.9.0-alpha.1+000.asdf123`. With this PR, it becomes `0.9.0-alpha.1` if the `asdf123` commit is the same.


consideRatio:
Manual testing concludes this is working as intended!

PR #57, opened 2019-10-18
--long flag to always output build information in image tags and chart version

consideRatio:
This PR adds the `--long` flag to always output build information in image tags and chart version. This mimics `git describe --long` that has does something very similar.

The `--long` flag could be useful to avoid producing duplicate entries when a build job is run twice, once because of a commit was pushed, and once again because of a tag was pushed and coupled with this commit. If `--long` us used with pushed commits, but not for pushed tags or with `--tag`, then we wouldn't risk pushing two charts of the same version for example.

chartpress-long-flag

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,6 +26,12 @@ script:
     cd zero-to-jupyterhub-k8s &&
     chartpress
   - git --no-pager diff
+
+  - chartpress --skip-build --tag 1.2.3-test.tag
+  - git --no-pager diff
+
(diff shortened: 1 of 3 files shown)


minrk:
It would be useful to be able to review PRs, especially before publishing releases if there aren't critical blocking issues in a broken release.

consideRatio:
Thanks for telling me, I need the feedback. Sorry for this, chartpress is due to my actions now in a bad state, caused technically by #

PR #63, opened 2019-10-20
support string images and items in lists

### LINKED ISSUE #10: Support images defined as string
Add support for images defined as `<registry>/<repo>:<tag>`.

Possible solutions to keep compatibility with images defined as mapping:

1. check the image value in `values.yaml`
2. if it is mapping use current way of setting the dictionary;
3. else use formatted string `"{registry}/{repo}:{tag}".format(...)`

Please let me know what you think and I will send a PR.

### LINKED ISSUE #39: Support extraContainers k8s container spec with valuesPath
I'm not sure if this is what chartpress is intended to do, but I'd like to try using chartpress to build an image defined in our jupyterhub.hub.extraContainers. extraContainers takes a list of dictionaries [where each has an `image` key](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/master/doc/source/advanced.md#hubextracontainers) with value of the form name:tag. However valuesPath will insert `name` and `tag` keys.

minrk:
Decision is based on the type of the field being set. If `valuesPath` points to a dict, set it the current `repository:, tag:` way, if it's a string, set a single `repository:tag` image string.

enables e.g. using chartpress to build images in `profileList` in zero-to-jupyterhub:

```yaml
charts:
  - name: jupyterhub

    images:
      default:
        valuesPath: jupyterhub.singleuser.profileList.0.kubespawner_override.image
      gap:
        valuesPath: jupyterhub.singleuser.profileList.1.kubespawner_override.image
```

closes #10
closes #39 (edit by @consideRatio)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,23 @@
 
 ## 0.4
 
+### 0.4.3
+
+- Support `valuesPath` pointing to a single `image:tag` string in
+  addition to a dict with separate `repository` and `tag` keys.
+- Support lists in `valuesPath` by using integer indices,
+  e.g. `section.list.1.image` for the yaml:
+
+  ```yaml
+  section:
+    list:
+        - first: item
+          image: "not set"
+        - second: item
+          image: "image:tag"  #  <--sets this here
+  ```
+
+
 ### 0.4.2
 
 - --long flag to always output build information in image tags and chart version [#57](https://github.com/jupyterhub/chartpress/pull/57) ([@consideRatio](https://github.com/consideRatio))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,10 +14,10 @@
   ```yaml
   section:
     list:
-        - first: item
-          image: "not set"
-        - second: item
-          image: "image:tag"  #  <--sets this here
+      - first: item
+        image: "not set"
+      - second: item
+        image: "image:tag"  #  <--sets this here
   ```
 
 
--- a/chartpress.py
+++ b/chartpress.py
@@ -277,6 +277,8 @@ def build_values(name, values_mods):
         values = yaml.load(f)
 
     for key, value in values_mods.items():
+        if not isinstance(value, dict) or set(value.keys()) != {'repository', 'tag'}:
+            raise ValueError(f"I only understand image updates with 'repository', 'tag', not: {value!r}")
         parts = key.split('.')
         mod_obj = parent = values
         for p in parts:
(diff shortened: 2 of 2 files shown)


chartpress.py
```diff
@@ -295,9 +299,12 @@ def build_values(name, values_mods):
                 )
 
             mod_obj['tag'] = value['tag']
+        elif isinstance(mod_obj, str) and set(value.keys()) == {'repository', 'tag'}:
```
consideRatio: the `build_images` function will return the `values_mods`, which will be used to get the `value` variable.

And this is how the `values_mods` are set:

```
        value_modifications[options['valuesPath']] = {
            'repository': image_name,
            'tag': SingleQuotedScalarString(image_tag),
        }
```

That makes me think that `set(value.keys()) == {'repository', 'tag'}` would always evaluate to true.

chartpress.py
```diff
@@ -295,9 +299,12 @@ def build_values(name, values_mods):
...
+            # scalar image string, not dict with separate repository, tag keys
+            parent[parts[-1]] = "{repository}:{tag}".format(**value)
```
consideRatio: This may fail if we want to update a numbered item in an array I think. We could use the variable `p` instead if we want to avoid that, or repeat this logic.

```python
            if p.isdigit():
                # integers are indices in lists
                p = int(p)
```

I don't see that use case at the moment though, but perhaps.

consideRatio: This was my quick check to figure out if you could reference a out-of-for-loop-scope variable.

image

CHANGELOG.md
```diff
@@ -4,6 +4,23 @@
...
+- Support lists in `valuesPath` by using integer indices,
+  e.g. `section.list.1.image` for the yaml:
+
+  ```yaml
+  section:
+    list:
+        - first: item
+          image: "not set"
+        - second: item
+          image: "image:tag"  #  <--sets this here
```
consideRatio: There are four spaces of indentation instead of two.

chartpress.py
minrk: indeed it will, for now! This was a protection against the fact that `values_mods` could be* a generic update to any fields in values, so may in the future contain updates other than image tags. But you are right that it does not right now, so this will always be true.

I updated the assumption to be explicit a little earlier, rather than checking it here, since I realized the dict-update also made the same assumption.

chartpress.py
minrk: Good catch! Fixed.

consideRatio:
I liked the helpful variable naming like `last_part`!

This LGTM, I'll leave it to you to merge if you consider it ready.

minrk:
Thanks for the review!

PR #66, opened 2019-10-22
fix latest tagged commit

minrk:
git describe shows current commit, not the tag commit, so the current commit was always being used, not the actual latest commit

use git rev-list to resolve the commit for the tag


consideRatio:
I reproduced the bug and concluded that this fixes it, I also understand the mistake i made:

```python
# it should be sha_for_commit_n_after_tag, which we don't care about
latest_tagged_commit = git_describe_head[2][1:]
```

Thanks for catching my mistake, LGTM!

PR #68, opened 2019-10-22
More thorough testing with pytest

### LINKED ISSUE #67: Setup more thorough pytest tests
# More thorough testing

Partially closed by #67.

## Managing testing content
There is a challenge with testing chartpress relating to its interaction with the git repository. For proper testing, we would need to make commits and tags to some git repository, within some branch.

The best idea that allows for these modifications etc is to work with a pytest fixture to create a new temp directory, copying testing content there, run git init etc, and add tags etc.

Discarded ideas:
- Having a separate repository with testing content etc that we would clone.
- Checking out a new branch, adding tags etc, and then discarding the branch and returning to previous branch.

## Testing content

```
README.md
chartpress.yaml
images/Dockerfile
charttest/values.yaml
charttest/Chart.yaml
```

## Tests
- [ ] Test --force-build
- [ ] Test --push
- [ ] Test --force-push
- [x] Test static and rendered dynamic values (TAG, LAST_COMMIT) etc are passed to the Dockerfile as --build-arg.
- [x] Test that the latest of a) latest tagged commit and b) latest modifying commit is used.
- [x] Test --version
- [x] Test --reset
- [x] Test --skip-build
- [x] Test --publish-chart
- [x] Test --extra-message
- ...

consideRatio:
Closes #67.

# PR Status
Ready for review -> merge.

## About test setup

The tests are setup by creating temporary git repositories where we copy a content. In the master branch of the temp git repo we put a test helm chart. In the gh-pages branch of the temp git repo we put a test helm chart repository containing the important index.yaml file.

Many tests work by capturing the output while running chartpress, and verifying that we find the right kind of statements in the output. It can be traces from chartpress itself, such as "Updating testchart/values.yaml with..." but also from docker such as "Successfully tagged..." or the `helm package` command etc.

## Added functionality: local chart

The `git_remote` function previously updated a passed "jupyterhub/helm-chart" string to be git@github.com:jupyterhub/helm-chart.git or similar. As I wanted to run tests without external dependencies etc, I made this function support being passed a `.` string without updating it, which allowed me to work directly with the gh-pages branch of the git repository chartpress is running on.

This also opens up the new possibility of not creating a dedicated helm-chart repo for the helm chart.

## bugfix: version bumping depending on what paths?
In this PR there is a slight change of behavior because of a bug fix.

```
chartpress.yaml
README.md
images/
- testimage
testchart/
- templates/
- values.yaml
- Chart.yaml
```

Previously, a change to this folder would give us a new version, now a change to the README.md wont. So it's not a big deal.

- For charts regarding version bumping, chartpress is considering the paths in chartpress.yaml, the helm charts folder, and chartpress.yaml, and any change to dependent images build context, Dockerfile or additional image paths.
- For images regarding version bumping, chartpress is considering the paths in chartpress.yaml, the build context, the Dockerfile, and chartpress.yaml.


chartpress.py
```diff
@@ -167,7 +172,6 @@ def image_needs_pushing(image):
         return False
 
 
-@lru_cache()
```
consideRatio: image_needs_building() changed during the duration of testing leading to an outdated value were used.

minrk: You can clear the cache by calling

```python
image_needs_rebuilding.cache_clear()
```

in the test fixture. That way, we don't lose this. We should also put this decorator on `latest_tag_or_mod_commit`.

consideRatio: None of these functions are pure though, hmmm... Do we really want to cache them at all?

I think the key question is: will chartpress, except for within a testing environment, ever risk reusing a cached value that no longer is relevant?

`latest_tag_or_mod_commit` can become outdated by git operations. No such operations are expected to be made during chartpress start and exit when used as a CLI, which is the duration of the cache as far as I understand. This function is currently called once for each chart's path, and once for each chart's images paths.

`image_needs_building` can become outdated by a docker build operation for example. It is called once for each image that is meant to be built from the build_images function.

`image_needs_pushing` can become outdated by a docker push operation for example, and is called TWICE, once to from image_needs_building and once if the `--push` flag is passed to verify it needs to be pushed (... and the tag isn't set? Adding tag implies `--force` to push and build currently. Hmm...)

---

What is your take considering this, should we cache? Also, can you confirm if I got it right that lru_cache will be reset in between calls to the chartpress CLI, where the process is shut down, but perhaps not if you have one process running that in turn has imported chartpress and use the functions directly without shutting down?

---

Do you think @lru_cache usage may relate to why travis-ci mess up with https://github.com/jupyterhub/chartpress/pull/68#issuecomment-545582996 ?

consideRatio:
The cache issue in https://github.com/jupyterhub/chartpress/pull/68#issuecomment-545582996 resolved, or perhaps rather stopped messing with the rendered output that made me thing something was wrong. I added back lru_cache and and started clearing the lru_cache when it had become outdated using the .clear_cache() function.

chartpress.py
consideRatio: (I updated the comment above to exclude no longer relevant discussion)

consideRatio:
Merge?

chartpress.py
consideRatio: While everything currently works, the tests will be a lot harder to get right and maintain with caching in place. I spent some extra hours on this but it doesn't doesn't do anything for the CLI except for image_needs_pushing cache in the specific circumstance that: an image was available locally, but not remotely, and we used `--push`.

Oh hmmmm, I can make a call to cache_clear alongside every call to chartpress during testing, it would make the test properly minimic the CLI which we want! I'm quite happy with this path as well!

consideRatio: I consider this resolved in ac85a6e

minrk:
This is excellent, great job!

Issue #61, opened 2019-10-19
PR Discussion: Let --publish-chart use --dependency-update on the helm package call


consideRatio:
@jhamman @scottyhq, I recall there was some issue with a dependency chart that didn't update. I imagine that it may be very relevant to have the `--publish-chart` flag make a call to `helm package` including the `--dependency-update` flag. Currently, we don't.

It probably makes sense to have this run as default within chartpress.

https://github.com/jupyterhub/chartpress/blob/84df258b335fe19d56d3fc849a9241d9c4eb7afe/chartpress.py#L370-L377

```
Usage:
  helm package [flags] [CHART_PATH] [...]

Flags:
      --app-version string   set the appVersion on the chart to this version
  -u, --dependency-update    update dependencies from "requirements.yaml" to dir "charts/" before packaging
  -d, --destination string   location to write the chart. (default ".")
  -h, --help                 help for package
      --key string           name of the key to use when signing. Used if --sign is true
      --keyring string       location of a public keyring (default "/home/erik/.gnupg/pubring.gpg")
      --save                 save packaged chart to local chart repository (default true)
      --sign                 use a PGP private key to sign this package
      --version string       set the version on the chart to this semver version
```

PR #69, opened 2019-10-24
Ensure to honor requirements.yaml before publishing a chart

consideRatio:
Closes https://github.com/jupyterhub/chartpress/issues/61.

In short, it makes sense to me to ensure that whatever is specified in the Helm chart's requirements.yaml is up to date with what is packaged.


minrk:
Makes sense!

PR #70, opened 2019-10-24
Add --force-build / --force-push, and don't let --tag imply it

### LINKED ISSUE #59: Feature proposal: a flag to enforce builds
Now that we check for local/remote existence of a image, we may end up failing to rebuild images if we want. I imagine a situation where we may want to get a fresher version of Ubuntu as base image for the hub image for example (`FROM ubuntu:18.04`), due to a security issue or similar, and that we want to re-push an update manually sometime.

I'm not sure.

consideRatio:
Closes #59.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,17 +42,21 @@ to build your chart(s) and image(s). Add `--push` to publish images to docker
 hub and `--publish-chart` to publish the chart and index to gh-pages.
 
 ```
-usage: chartpress [-h] [--push] [--publish-chart]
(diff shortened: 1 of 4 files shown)


minrk:
This looks great to me. Do you want to add a test for it, now that #68 is merged?

consideRatio:
Thanks! I'll add a test as well!

consideRatio:
@minrk quickly added a test for --force-build, but not for --force-push. There is currently none for --push either and it is documented in [tests/README.md](tests/README.md).

PR #75, opened 2019-11-16
[WIP] Preliminary check of chart publication

consideRatio:
I'm not sure what kind of usage experience we should aim for, but I know that I want to have chartpress help me perform checks if a helm chart has been published already. And I want to have this test be usuable to allow a CI system to skip tests if it wants to based on that, as well as be able to let chartpress only publish a chart if it isn't already published.

For example, in z2jh, when we don't make a change that affects the Helm chart, because we update the documentation only or similar, we don't need to run tests of the Helm chart. How would we like to use chartpress in those circumstances?

Here are some ideas of usage.

```shell
# stop early if chart version is published
# ...
# exit with what status code btw?
chartpress --exit-if-published

# print any published information about the chart if
# there is any, and exit like --version does after
chartpress --published-info
```

@sgibson91 @betatim @minrk, do you have ideas on how we design this kind of feature to be a useful and user friendly feature?

### DIFF
--- a/chartpress.py
+++ b/chartpress.py
@@ -322,7 +322,11 @@ def _get_published_chart_information(chart_repo_url, chart_name, chart_version):
     if not chart_releases:
         return None
 
-    matching_chart_release = dict(next((r for r in chart_releases if r["version"]==chart_version), {}))
+    matching_chart_release = None
+    for release in chart_releases:
+         if release["version"] == chart_version:
+            matching_chart_release = release
+            break
     return matching_chart_release
 
 
--- a/chartpress.py
+++ b/chartpress.py
@@ -11,9 +11,11 @@ from functools import lru_cache, partial
 import os
 import pipes
 import re
+import requests
 import shutil
 import subprocess
 from tempfile import TemporaryDirectory
+from urllib.parse import urljoin
 
 import docker
 from ruamel.yaml import YAML
@@ -290,6 +292,38 @@ def _strip_identifiers_build_suffix(identifier):
     # . or - followed by three or more digits, a dot, an commit sha of four
     # or more alphanumeric characters.
     return re.sub(r'[-\.]\d{3,}\.\w{4,}\Z', "", identifier)
+
+
+def _get_published_chart_information(chart_repo_url, chart_name, chart_version):
+    """
+    Provide information about a chart release from a chart repository url along
+    with a chart name and version to search for.
+
+    - Returns the dictionary representing the matching chart entry from the Helm
+      chart repository if a matching entry is found.
+    - Returns None if a Helm chart repostiory is found but the matching chart
+      name and version isn't.
+    - Raises errors if the Helm chart repository isn't accessible or doesn't
+      seem to be one.
+    """
+    index_url = urljoin(chart_repo_url + "/", "index.yaml")
+    response = requests.get(index_url)
+
+    # this can throw errors, and doing that makes sense
+    # - connection failure
+    # - response doesn't seem like a Helm chart repository
+    response.raise_for_status()
+    index = yaml.load(response.content)
+    chart_repository_entries = index["entries"]
+
+    # let's not throw errors now, but instead return None if
+    # we fail to find a matching chart version
+    chart_releases = chart_repository_entries.get(chart_name, None)
+    if not chart_releases:
+        return None
+
+    matching_chart_release = dict(next((r for r in chart_releases if r["version"]==chart_version), {}))
+    return matching_chart_release
 
 
 def build_images(prefix, images, tag=None, push=False, force_push=False, chart_version=None, force_build=False, skip_build=False, long=False):
@@ -662,6 +696,11 @@ def main(args=None):
         action='store_true',
         help='Print current chartpress version and exit.',
     )
+    argparser.add_argument(
+        '--exit-if-published',
+        action='store_true',
+        help='Exit after chart version has been determined if that chart version is already published.',
+    )
 
     argparser.add_argument(
         '--commit-range',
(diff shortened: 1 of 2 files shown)


betatim:
I'll start with a short summary/repeating the idea back to see if I got it right:

The goal is to not build and publish a new helm chart when the changes in the repository are only to the documentation.

In BinderHub [this line](https://github.com/jupyterhub/binderhub/blob/c80b26cafe73f253428a3bebfe4292dfb76796e3/ci/test-helm#L51) is what runs chartpress and causes it to publish a new version (I think). I am not sure I fully understand how we aren't publishing a new chart for each commit in a PR :-/ However we aren't so something is catching that already.

We also had a bug once that `chartpress` was determining that there was nothing to do because we didn't include all the directories in the repo that could cause a change to the chart.

My take away from those two points is that there are already mechanisms to prevent building(?) and publishing a chart when it isn't needed. As well as me not fully understanding how they work. This makes it a bit tricky to think about this PR.

In principle not republishing gets a 👍  however I also think we shouldn't be republishing if everything is working as designed and we use the tool as intended. How do we figure out if this is true?

consideRatio:
> The goal is to not build and publish a new helm chart when the changes in the repository are only to the documentation.

Yepp!

> We also had a bug once that chartpress was determining that there was nothing to do because we didn't include all the directories in the repo that could cause a change to the chart.

I've fixed some bugs regarding image paths etc, they are part of chartpress==0.4.3, it comes from this commit https://github.com/jupyterhub/chartpress/commit/655223ec95ae841cc3eed56e9e5cdc2a5a2508ff. [BinderHub currently use chartpress==0.3.1](https://github.com/jupyterhub/binderhub/blob/master/dev-requirements.txt#L9). I'll submit a PR to fix some things because there are various outdated aspects of how chartpress is used I see.

This isn't only about not republishing, it is about speeding up builds for PRs not affecting the chart itself by not running the full test suite all the time when it hasn't changed. With this in mind, how would a bash script be best informed by chartpress if the chart of this state is already published? Through passing an error? By writing text output which we grep or similar? Hmmm...

consideRatio:
## Notes for self

I've learned from using grep, that you can do for example.

```
echo "this is a test string with the secret keyword meow" | grep "meow"
if [ $? -eq 0 ]; then
  echo "meow was ushered!"
else
  echo "meow was NOT ushered"
fi
```

I think it could be nice to use that kind of system, where we would do.

```
chartpress --is-published
if [ $? -eq 0 ]; then
  echo "the chart version identified from the current git-repo state is already published"
else
  echo "the chart version identified from the current state is not already published!"
fi
```

Also, there are two ways to check if a chart is published.
1. To use a web-url to a helm chart repo's index.yaml file, as we typically have made explicit in chartpress.yaml's `repo.published` field for a chart, and look for the version in the helm chart repo index.
2. We extract the `index.yaml` file from the gh-pages branch of the git repo itself and look for it.

I can see the need for choosing among these options, but if we choose to implement only one, i say #1 is the most important, but for more private stuff, it is nice to inspect a git repo instead... Hmm...arrgh nah option 1 is the way to go for now, especially since i realize we can manage access to index.yaml for helm like [described here](https://blog.softwaremill.com/hosting-helm-private-repository-from-github-ff3fa940d0b7) not using gh-pages, hmm...

---

Implementation ideas based on this.

```
chartpress --published-info
# prints some info and exit with error code 1 if it isnt, and 0 if it is

chartpress --force-publish
# will ignore if it is already published or not

chartpress --publish
# will print that it didn't publish if it was already published
```

chartpress.py
```diff
@@ -292,6 +294,38 @@ def _strip_identifiers_build_suffix(identifier):
...
+    with a chart name and version to search for.
+
+    - Returns the dictionary representing the matching chart entry from the Helm
+      chart repository if a matching entry is found.
+    - Returns None if a Helm chart repostiory is found but the matching chart
+      name and version isn't.
+    - Raises errors if the Helm chart repository isn't accessible or doesn't
+      seem to be one.
+    """
+    index_url = urljoin(chart_repo_url + "/", "index.yaml")
```
minrk: urljoin is what adds the slash here

```suggestion
    index_url = urljoin(chart_repo_url, "index.yaml")
```

chartpress.py
```diff
@@ -292,6 +294,38 @@ def _strip_identifiers_build_suffix(identifier):
...
+    index = yaml.load(response.content)
+    chart_repository_entries = index["entries"]
+
+    # let's not throw errors now, but instead return None if
+    # we fail to find a matching chart version
+    chart_releases = chart_repository_entries.get(chart_name, None)
+    if not chart_releases:
+        return None
+
+    matching_chart_release = dict(next((r for r in chart_releases if r["version"]==chart_version), {}))
```
minrk: What's this `dict` call doing?

Perhaps clearer:

```suggestion
    matching_chart_release = None
    for release in chart_releases:
         if release["version"] == chart_version:
            matching_chart_release = release
            break
```

chartpress.py
```diff
@@ -662,6 +696,11 @@ def main(args=None):
         action='store_true',
         help='Print current chartpress version and exit.',
     )
+    argparser.add_argument(
+        '--exit-if-published',
```
minrk: Do we want this to be the default and opt-out, rather than opt-in? i.e. `--overwrite-chart` and error if already published by default instead?

consideRatio: Well I've considered the API that makes sense to implement, and what i concluded isn't reflected in this PR's code yet but only in this comment: https://github.com/jupyterhub/chartpress/pull/75#issuecomment-559023589.

I currently think that overriding charts in a helm chart repo should be considered opt-in, and not be done by default.

chartpress.py
consideRatio: I was very confused by this as well. I still don't get this properly. But, if I remove it my tests fail when this function is called, and here is some debugging information of relevance.

Adding some print statements:
```python
    print(chart_repo_url)
    # https://jupyterhub.github.io/helm-chart

    print(urljoin(chart_repo_url, "index.yaml"))
    # https://jupyterhub.github.io/index.yaml

    print(urljoin(chart_repo_url + "/", "index.yaml"))
    # https://jupyterhub.github.io/helm-chart/index.yaml
```

consideRatio:
```ipython
In [4]: urllib.parse.urljoin("https://example.local/ignored_subpath", "index.yaml")
Out[4]: 'https://example.local/index.yaml'

In [5]: urllib.parse.urljoin("https://example.local/not_ignored_subpath/", "index.yaml")
Out[5]: 'https://example.local/not_ignored_subpath/index.yaml'

In [6]: urllib.parse.urljoin("https://example.local/not_ignored_subpath//", "index.yaml")
Out[6]: 'https://example.local/not_ignored_subpath/index.yaml'
```

PR #85, opened 2019-11-30
Update changelog for 0.5.0

consideRatio:
I'm not 100% on this, should I bump this to 0.5.0 because there are new features and minor tweaks of behavior, or make it 0.4.4 as it mainly fixes bugs and weird behavior?

This is what makes me thing it should be 0.5.0:
- Add --force-build / --force-push, and don't let --tag imply it [#70](https://github.com/jupyterhub/chartpress/pull/70) ([@consideRatio](https://github.com/consideRatio))
- `helm dependency update` is now run as part of publishing, this ensures we honor requirements.yaml before publishing a chart [#69](https://github.com/jupyterhub/chartpress/pull/69) ([@consideRatio](https://github.com/consideRatio))

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,28 @@
 
 ## Unreleased
 
-## 0.4
+## [0.5]
 
-### 0.4.3 (Breaking changes)
+### [0.5.0]
+
+#### Added
+
+- Add --force-build / --force-push, and don't let --tag imply it [#70](https://github.com/jupyterhub/chartpress/pull/70) ([@consideRatio](https://github.com/consideRatio))
+- `helm dependency update` is now run as part of publishing, this ensures we honor requirements.yaml before publishing a chart [#69](https://github.com/jupyterhub/chartpress/pull/69) ([@consideRatio](https://github.com/consideRatio))
+
+#### Fixes
+
+- Fix regression to make a chart's `images` configuration optional again [#77](https://github.com/jupyterhub/chartpress/pull/77) ([@jacobtomlinson](https://github.com/jacobtomlinson))
+- Fix regarding image paths as part of setting up thorough testing with PyTest [#68](https://github.com/jupyterhub/chartpress/pull/68) ([@consideRatio](https://github.com/consideRatio))
+
+#### Maintenance
+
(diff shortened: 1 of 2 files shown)


consideRatio:
@bitnik and @betatim making a release of chartpress is what fixes the behavior which I though was already fixed in binderhub that caused the need for https://github.com/jupyterhub/binderhub/pull/1014, as @bitnik figured out in https://github.com/jupyterhub/binderhub/pull/1004#discussion_r351630260.

---

I'd like help making a call on if this should be considered 0.5.0 or 0.4.4, and a OK check to release.

manics:
I haven't looked in detail, but based on the changelog it sounds like 0.4.3 contained breaking changes so should've been tagged as 0.5.0? If that's the case I vote for 0.5.0, and consider marking the whole 0.4.x series broken.

consideRatio:
0.4.3 is not broken, but 0.4.0-0.4.2 is introduced an idea that had to be reverted entirely into what became 0.4.3. So i felt like 0.4.3 was what 0.4.0 was meant to be that finally hot it working.

Between 0.4.3 and this, it is some changes to behavior but not huge... hmmmm... i recall major is for breaking changes after we havr a 1.0.0, but before that, how should minor patch be used with regards to breaking/new features/fixes? If we cant yet consider using major for breaking only, that better be for minor, and patch could be allowed to contain some minor new features perhaps?

Hmmmm

betatim:
I would go with 0.5.0.

For me a switch from 0.x.y to 0.x.(y+1) means "you can update this without even reading the changelog, it will just work. It for sure won't lead to breakage in your usage. Updating is a no brainer." However it sounds here we want people to look at the changes and engage their brain. So 0.5.0 it is.

The number of users of chartpress is small but those who do use it use it as part of their release/deployment setup. For me the tools that I use as part of releasing and deployment (to production systems) should change slowly and any change always needs thinking about. Having to take time to rework release/deployment tooling is inconvenient (because you expect breakage) but is 100 times better than getting surprised by something having broken because you upgraded something (potentially weeks ago).

So incurring a little extra work because "oh wow they went from 0.4.x to 0.5.0 better check why" is better than people not paying attention and then getting caught by something that stopped working when they deploy/release a project (which is already a hectic time).

---

SemVer for pre 1.0.0 says: ["all bets are off, anything goes"](https://semver.org/#spec-item-4).

PR #87, opened 2019-12-17
Prefix build info with n and h to ensure SemVer 2 validity to solve Helm 3 compatibility

### LINKED ISSUE #86: New version format doesn't work with helm 3
Testing with the latest binderhub helm chart dev releases:
```
$ helm version`
version.BuildInfo{Version:"v3.0.0", GitCommit:"e29ce2a54e96cd02ccfce88bee4f58bb6e2a28b6", GitTreeState:"clean", GoVersion:"go1.13.4"}
```
```
$ helm repo update
Hang tight while we grab the latest from your chart repositories...
...Successfully got an update from the "jupyterhub" chart repository
...Successfully got an update from the "stable" chart repository
Update Complete. ⎈ Happy Helming!⎈ 
```

This works:
```
helm upgrade binder-test jupyterhub/binderhub --install --version=0.2.0-c04966f --namespace=binder-test -f secret.yaml -f config.yaml
```

This fails:
```
helm upgrade binder-test jupyterhub/binderhub --install --version=0.2.0-028.9ba1fc3 --namespace=binder-test -f secret.yaml -f config.yaml
Error: failed to download "jupyterhub/binderhub" (hint: running `helm repo update` may help)
```

This also fails:
```
$ cd helm-chart/binderhub/
$ cat requirements.yaml 
dependencies:
- name: jupyterhub
  version: "0.9.0-alpha.1.071.b0d70c2"
  repository: "https://jupyterhub.github.io/helm-chart"
$ helm dependency update
Hang tight while we grab the latest from your chart repositories...
...Successfully got an update from the "jupyterhub" chart repository
...Successfully got an update from the "stable" chart repository
Update Complete. ⎈Happy Helming!⎈
Error: dependency "jupyterhub" has an invalid version/constraint format: constraint Parser Error
```

consideRatio:
Fixes #86 with 14f44dc + test updates.

In short, we now make `0.9.0-alpha.1.095.123asdf` become `0.9.0-alpha.1.n095.h123asdf` to ensure we never have a numerical identifier with a leading zero, which is breaking Helm 3 usage of our chart because its dependency masterminds/semver now does a better job verifying if the version is a valid semver version etc.

I failed to spot the suggested version with a leading zero in a fully numerical identifier wasn't a valid semver version before =/

This will need a minor version bump for a release, and could once again cause issues from chartpress that @henchc @betatim has helped worked through in the past for binderhub and mybinder.org.

### DIFF
--- a/chartpress.py
+++ b/chartpress.py
@@ -251,22 +251,24 @@ def _get_identifier(tag, n_commits, commit, long):
 
     This function should provide valid Helm chart versions, which means they
     need to be valid SemVer 2 version strings. It also needs to return valid
-    image tags, which means they need to not contain `+` signs either.
+    image tags, which means they need to not contain `+` signs either. We prefix
+    with n and h to ensure we don't get a numerical hash starting with 0 because
+    those are invalid SemVer 2 versions.
 
     Example:
         tag="0.1.2", n_commits="5", commit="asdf1234", long=True,
-        should return "0.1.2-005.asdf1234".
+        should return "0.1.2-n005.hasdf1234".
     """
     n_commits = int(n_commits)
 
     if n_commits > 0 or long:
         if "-" in tag:
             # append a pre-release tag, with a . separator
-            # 0.1.2-alpha.1 -> 0.1.2-alpha.1.n.sha
+            # 0.1.2-alpha.1 -> 0.1.2-alpha.1.n.h
             return f"{tag}.{n_commits:03d}.{commit}"
         else:
             # append a release tag, with a - separator
-            # 0.1.2 -> 0.1.2-n.sha
+            # 0.1.2 -> 0.1.2-n.h
             return f"{tag}-{n_commits:03d}.{commit}"
     else:
         return f"{tag}"
@@ -275,12 +277,12 @@ def _get_identifier(tag, n_commits, commit, long):
 def _strip_identifiers_build_suffix(identifier):
     """
     Return a stripped chart version or image tag (identifier) without its build
-    suffix (.005.asdf1234), leaving it to represent a Semver 2 release or
+    suffix (.n005.hasdf1234), leaving it to represent a Semver 2 release or
     pre-release.
 
     Example:
-        identifier: "0.1.2-005.asdf1234"            returns: "0.1.2"
-        identifier: "0.1.2-alpha.1.005.asdf1234"    returns: "0.1.2-alpha.1"
+        identifier: "0.1.2-n005.hasdf1234"            returns: "0.1.2"
+        identifier: "0.1.2-alpha.1.n005.hasdf1234"    returns: "0.1.2-alpha.1"
     """
     # split away official SemVer 2 build specifications if used
     if "+" in identifier:
@@ -322,11 +324,11 @@ def build_images(prefix, images, tag=None, push=False, force_push=False, chart_v
 
         Example 1:
         - long=False: 0.9.0
-        - long=True:  0.9.0-000.asdf1234
+        - long=True:  0.9.0-n000.hasdf1234
 
         Example 2:
-        - long=False: 0.9.0-004.sdfg2345
-        - long=True:  0.9.0-004.sdfg2345
+        - long=False: 0.9.0-n004.hsdfg2345
+        - long=True:  0.9.0-n004.hsdfg2345
     """
     value_modifications = {}
     for name, options in images.items():
@@ -459,11 +461,11 @@ def build_chart(name, version=None, paths=None, long=False):
 
     Example versions constructed:
         - 0.9.0-alpha.1
-        - 0.9.0-alpha.1.000.asdf1234 (--long)
-        - 0.9.0-alpha.1.005.sdfg2345
-        - 0.9.0-alpha.1.005.sdfg2345 (--long)
+        - 0.9.0-alpha.1.n000.hasdf1234 (--long)
+        - 0.9.0-alpha.1.n005.hsdfg2345
+        - 0.9.0-alpha.1.n005.hsdfg2345 (--long)
         - 0.9.0
-        - 0.9.0-002.dfgh3456
+        - 0.9.0-n002.hdfgh3456
     """
     chart_file = os.path.join(name, 'Chart.yaml')
     with open(chart_file) as f:
--- a/chartpress.py
+++ b/chartpress.py
@@ -265,11 +265,11 @@ def _get_identifier(tag, n_commits, commit, long):
         if "-" in tag:
             # append a pre-release tag, with a . separator
             # 0.1.2-alpha.1 -> 0.1.2-alpha.1.n.h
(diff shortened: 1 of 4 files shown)


minrk:
Arg, that's really annoying. These are getting really long, so maybe replace `sha` with just `h` for hash? It's not relevant that it's a sha hash, and this is also what conda uses in build strings.

consideRatio:
Yes, very long, :/

Im afraid of the one letter solution, as that would masquerade it into the hash. Git use `g` prefix also btw.

Im okay with various options, perhaps even more okay with accepting bad luck can happen and writing out `invalid` or similar when we get a pure numerical sha starting with zero. I dislike having a single letter the most i think, it made me really confused with the git describe prefixed `g` for example

manics:
Consistency is more important than what looks nice. If the value can sometimes be `invalid` this means no tooling can rely on it's value, in which case you might as well drop it and use something more readable such as a timestamp (with a letter prefix 😀).

I don't think `h` vs `g` vs `sha` makes much difference to the readability since it all gets merged with the hash anyway e.g. `shaabc1234` vs `habc1234`. OTOH given the current proposed string is 29 chars shortening it to 27 isn't that big a saving. Given the length I presume everyone will always copy-and-paste these strings.

Any of `h`, `g`, `sha`, `z`, sound fine to me, and focus on ensuring correctness so another change isn't necessary.

minrk:
Yeah, agreed that validity is most important. It’s just frustrating since human intelligibility was the primary motivation that started this whole change, and it’s starting to look a lot less human-friendly in the end. Let’s go with just ‘h’ and release.

consideRatio:
Anyone wanting to replace the `sha` prefix with `h` are welcome to do so by pushing to my branch for this pr. I only have a mobile until 13th jan but i can review changes!

chartpress.py
```diff
@@ -255,32 +255,32 @@ def _get_identifier(tag, n_commits, commit, long):
...
+        should return "0.1.2-n005.shaasdf1234".
     """
     n_commits = int(n_commits)
 
     if n_commits > 0 or long:
         if "-" in tag:
             # append a pre-release tag, with a . separator
             # 0.1.2-alpha.1 -> 0.1.2-alpha.1.n.sha
-            return f"{tag}.{n_commits:03d}.{commit}"
+            return f"{tag}.n{n_commits:03d}.sha{commit}"
```
betatim:
```suggestion
            return f"{tag}.n{n_commits:03d}.h{commit}"
```

chartpress.py
```diff
@@ -255,32 +255,32 @@ def _get_identifier(tag, n_commits, commit, long):
...
         else:
             # append a release tag, with a - separator
             # 0.1.2 -> 0.1.2-n.sha
-            return f"{tag}-{n_commits:03d}.{commit}"
+            return f"{tag}-n{n_commits:03d}.sha{commit}"
```
betatim:
```suggestion
            return f"{tag}-n{n_commits:03d}.h{commit}"
```

chartpress.py
```diff
@@ -255,32 +255,32 @@ def _get_identifier(tag, n_commits, commit, long):
...
     else:
         return f"{tag}"
 
 
 def _strip_identifiers_build_suffix(identifier):
     """
     Return a stripped chart version or image tag (identifier) without its build
-    suffix (.005.asdf1234), leaving it to represent a Semver 2 release or
+    suffix (.n005.shaasdf1234), leaving it to represent a Semver 2 release or
```
betatim:
```suggestion
    suffix (.n005.hasdf1234), leaving it to represent a Semver 2 release or
```

chartpress.py
```diff
@@ -255,32 +255,32 @@ def _get_identifier(tag, n_commits, commit, long):
...
     pre-release.
 
     Example:
-        identifier: "0.1.2-005.asdf1234"            returns: "0.1.2"
-        identifier: "0.1.2-alpha.1.005.asdf1234"    returns: "0.1.2-alpha.1"
+        identifier: "0.1.2-n005.shaasdf1234"            returns: "0.1.2"
```
betatim:
```suggestion
        identifier: "0.1.2-n005.hasdf1234"            returns: "0.1.2"
```

chartpress.py
```diff
@@ -255,32 +255,32 @@ def _get_identifier(tag, n_commits, commit, long):
...
+        identifier: "0.1.2-alpha.1.n005.shaasdf1234"    returns: "0.1.2-alpha.1"
```
betatim:
```suggestion
        identifier: "0.1.2-alpha.1.n005.hasdf1234"    returns: "0.1.2-alpha.1"
```

chartpress.py
```diff
@@ -289,7 +289,7 @@ def _strip_identifiers_build_suffix(identifier):
     # split away our custom build specification: something ending in either
     # . or - followed by three or more digits, a dot, an commit sha of four
     # or more alphanumeric characters.
-    return re.sub(r'[-\.]\d{3,}\.\w{4,}\Z', "", identifier)
+    return re.sub(r'[-\.]n\d{3,}\.sha\w{4,}\Z', "", identifier)
```
betatim:
```suggestion
    return re.sub(r'[-\.]n\d{3,}\.h\w{4,}\Z', "", identifier)
```

chartpress.py
```diff
@@ -322,11 +322,11 @@ def build_images(prefix, images, tag=None, push=False, force_push=False, chart_v
 
         Example 1:
         - long=False: 0.9.0
-        - long=True:  0.9.0-000.asdf1234
+        - long=True:  0.9.0-n000.shaasdf1234
```
betatim:
```suggestion
        - long=True:  0.9.0-n000.hasdf1234
```

chartpress.py
```diff
@@ -322,11 +322,11 @@ def build_images(prefix, images, tag=None, push=False, force_push=False, chart_v
...
 
         Example 2:
-        - long=False: 0.9.0-004.sdfg2345
-        - long=True:  0.9.0-004.sdfg2345
+        - long=False: 0.9.0-n004.shasdfg2345
```
betatim:
```suggestion
        - long=False: 0.9.0-n004.hsdfg2345
```

chartpress.py
```diff
@@ -322,11 +322,11 @@ def build_images(prefix, images, tag=None, push=False, force_push=False, chart_v
...
+        - long=True:  0.9.0-n004.shasdfg2345
```
betatim:
```suggestion
        - long=True:  0.9.0-n004.hsdfg2345
```

betatim:
I started replacing `sha` with `h` but then realised there are far too many instances to do this via GitHub suggestions, so I stopped :-/

consideRatio:
Thanks @betatim for the attempt! I'm home now at my computer and stepped through the commits and made a rebased update.

consideRatio:
I made a 0.6.0 release with this!

PR #90, opened 2020-03-12
support overriding imageName

minrk:
Allows building individual images without the prefix+name structure

in nbviewer, I want to build `jupyter/nbviewer` and `jupyter/nbviewer-statuspage`, but use `jupyter/nbviewer-` as prefix.

I know I *can* use `jupyter/` as prefix and hve image names `nbviewer`, `nbviewer-statuspage`, but this seems like something that might come up again.

### DIFF
--- a/README.md
+++ b/README.md
@@ -164,9 +164,12 @@ charts:
     # images to build for this chart (optional)
     images:
       binderhub:
-        # Template docker build arguments to be passed using docker's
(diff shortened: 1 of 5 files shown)


consideRatio:
Thanks @minrk! I added some documentation and a test for this as well. I'll make a release as soon as I find some time.

PR #96, opened 2020-09-08
add chartpress --list-images

### LINKED ISSUE #94: --push-latest-tag by default
### Proposed change
To publish a `latest` image tag by default and make it configurable by a flag.

### Who would use this feature?
It would allow for example a GitHub action setup to do a security scan of the latest image be able to do it without having a lot of code to extract the latest tagged image using chartpress or similar. This is a discussion arising from https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1712#issuecomment-656746691.

### (Optional): Suggest a solution
We add a boolean flag which defaults to something that makes us also push `latest` image tags when we `--push` the images. I suggest `--push-latest-tag=false` as the override of this default or perhaps `--no-push-latest-tag` or something like that.

minrk:
outputs list of images to stdout. Listing images implies skip-build.

for convenient consumption of the list of images, e.g. to pass to trivy for security scans

To make chartpress output pipeable, send all log messages to stderr instead of stdout, as is standard practice.

I think this closes #94 unless I misunderstood the goal there.

Since this does move prevous output to stderr, if anyone was relying on chartpress stdout, this could be considered a backward-incompatible change. But since our output was not really machine-readable, I'd call this a minor revision.

### DIFF
--- a/chartpress.py
+++ b/chartpress.py
@@ -49,7 +49,11 @@ def run_cmd(call, cmd, *, echo=True, **kwargs):
     return call(cmd, **kwargs)
 
 
-check_call = partial(run_cmd, subprocess.check_call, stdout=sys.stderr)
+def check_call(cmd, **kwargs):
+    kwargs.setdefault("stdout", sys.stderr.fileno())
+    return run_cmd(subprocess.check_call, cmd, **kwargs)
+
+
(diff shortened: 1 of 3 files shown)


yuvipanda:
Thanks, @minrk!

PR #102, opened 2020-11-14
Add --force-publish-chart and default to not overwriting

### LINKED ISSUE #60: Feature proposal: check if chart is published
It could be good to have a flag to check if a chart of the version that would be set is already available in the associated helm chart repository I think. That way, we could have the means to resolve this issue for example: https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1451.

### LINKED ISSUE #73: PR Discussion: require --force-publish-chart to override charts during --publish-charts
In https://github.com/jupyterhub/chartpress/pull/68/commits/ccbbba71d4419a2da14de6bc0287deca63de5110 I documented the current procedure for --publish-chart.

We currently act as if we have a `--force-publish-chart` flag set by default when we use `--publish-chart`. I think we should require `--force-publish-chart` to be passed to get the current behavior where a new chart of the same version will override an existing chart.

This relates to #60 that is about making chartpress provide a easy way to check if a chart is already published before running tests etc again if nothing has changed.

consideRatio:
This feature help us avoid publishing the same chart version twice unless `--force-publish-chart` is passed. This mirrors the `--push` and `--force-push` behavior of docker images.

There are several reasons to not push an already existing chart:
1. It can trigger other pointless actions, such as henchbot suggesting mybinder.org-deploy to update.
2. It will clutter the gh-pages branch of the helm-chart git repository we push to. This clutter is both visual for any user browsing the versions and makes the helm chart repo grow bigger and bigger.

Closes #60 (the need for this kind of feature)
Closes #73 (implementation planning)
Closes #75 (an old PR to do this pretty much)

## Review notes
This is the relevant implementation changing the default behavior to check unless --force-publish-chart has been passed.

https://github.com/jupyterhub/chartpress/pull/102/files#diff-b221c94b491bd2ffe6ebc5416067fd58b378f4eda722e75754468ed9f24aefefR587-R601

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,9 +86,9 @@ hub and `--publish-chart` to publish the chart and index to gh-pages.
 
 ```
 usage: chartpress [-h] [--push] [--force-push] [--publish-chart]
-                  [--extra-message EXTRA_MESSAGE] [--tag TAG | --long]
-                  [--image-prefix IMAGE_PREFIX] [--reset]
-                  [--skip-build | --force-build] [--version]
+                  [--force-publish-chart] [--extra-message EXTRA_MESSAGE]
+                  [--tag TAG | --long] [--image-prefix IMAGE_PREFIX] [--reset]
+                  [--skip-build | --force-build] [--list-images] [--version]
                   [--commit-range COMMIT_RANGE]
 
 Automate building and publishing helm charts and associated images. This is
@@ -101,8 +101,14 @@ optional arguments:
   --force-push          Push built images to their image registries,
                         regardless if it would replace an existing image.
   --publish-chart       Package a Helm chart and publish it to a Helm chart
-                        repository contructed with a GitHub git repository and
-                        GitHub pages.
+                        registry constructed with a GitHub git repository and
+                        GitHub pages, but not if it would replace an existing
+                        chart version.
+  --force-publish-chart
+                        Package a Helm chart and publish it to a Helm chart
+                        registry constructed with a GitHub git repository and
+                        GitHub pages, regardless if it would replace an
+                        existing chart version
   --extra-message EXTRA_MESSAGE
                         Extra message to add to the commit message when
                         publishing charts.
@@ -119,6 +125,8 @@ optional arguments:
   --skip-build          Skip the image build step.
   --force-build         Enforce the image build step, regardless of if the
                         image already is available either locally or remotely.
+  --list-images         print list of images to stdout. Images will not be
+                        built.
   --version             Print current chartpress version and exit.
   --commit-range COMMIT_RANGE
                         Deprecated: this flag will be ignored. The new logic
--- a/chartpress.py
+++ b/chartpress.py
@@ -70,9 +70,8 @@ def git_remote(git_repo):
 
     github_token = os.getenv(GITHUB_TOKEN_KEY)
     if github_token:
-        return 'https://{0}@github.com/{1}'.format(
-            github_token, git_repo)
-    return 'git@github.com:{0}'.format(git_repo)
+        return f'https://{github_token}@github.com/{git_repo}'
+    return f'git@github.com:{git_repo}'
 
 
 @lru_cache()
@@ -206,7 +205,7 @@ def build_image(image_spec, context_path, dockerfile_path=None, build_args=None)
     if dockerfile_path:
         cmd.extend(['-f', dockerfile_path])
     for k, v in (build_args or {}).items():
-        cmd += ['--build-arg', '{}={}'.format(k, v)]
+        cmd += ['--build-arg', f'{k}={v}']
     check_call(cmd)
 
 @lru_cache()
@@ -383,7 +382,7 @@ def build_images(prefix, images, tag=None, push=False, force_push=False, chart_v
             ).decode('utf-8').strip()
             image_tag = _get_identifier(chart_version, n_commits, image_commit, long)
         image_name = options.get('imageName', prefix + name)
-        image_spec = '{}:{}'.format(image_name, image_tag)
+        image_spec = f'{image_name}:{image_tag}'
 
         values_path_list = options.get('valuesPath', [])
         if isinstance(values_path_list, str):
@@ -537,7 +536,7 @@ def build_chart(name, version=None, paths=None, long=False):
     return version
 
 
-def publish_pages(chart_name, chart_version, chart_repo_github_path, chart_repo_url, extra_message=''):
+def publish_pages(chart_name, chart_version, chart_repo_github_path, chart_repo_url, extra_message="", force=False):
     """
     Update a Helm chart registry hosted in the gh-pages branch of a GitHub git
     repository.
@@ -546,13 +545,15 @@ def publish_pages(chart_name, chart_version, chart_repo_github_path, chart_repo_
 
     1. Clone the Helm chart registry as found in the gh-pages branch of a git
        reposistory.
-    2. Create a temporary directory and `helm package` the chart into a file
+    2. If --force-publish-chart isn't specified, then verify that we won't
+       overwrite an existing chart version.
+    3. Create a temporary directory and `helm package` the chart into a file
        within this temporary directory now only containing the chart .tar file.
-    3. Generate a index.yaml with `helm repo index` based on charts found in the
+    4. Generate a index.yaml with `helm repo index` based on charts found in the
        temporary directory folder (a single one), and then merge in the bigger
        and existing index.yaml from the cloned Helm chart registry using the
        --merge flag.
-    4. Copy the new index.yaml and packaged Helm chart .tar into the gh-pages
+    5. Copy the new index.yaml and packaged Helm chart .tar into the gh-pages
        branch, commit it, and push it back to the origin remote.
 
     Note that if we would add the new chart .tar file next to the other .tar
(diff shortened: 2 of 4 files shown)


chartpress.py
```diff
@@ -566,20 +567,37 @@ def publish_pages(chart_name, chart_version, chart_repo_github_path, chart_repo_
...
+            [
+                'git', 'clone', '--no-checkout',
+                git_remote(chart_repo_github_path),
+                checkout_dir,
+            ],
+            # warning: if echoed, this call could reveal the github token
+            echo=True,
+        )
+    else:
+        check_call(['git', 'fetch'], cwd=checkout_dir, echo=True)
```
consideRatio, line 584: `chartpress --publish-chart` leads to a directory being created. If you try to publish the same chart version twice with an existing directory you get an error with `git clone`. This avoids that by reusing the previous git clone if it exists already.

README.md
```diff
@@ -119,6 +125,8 @@ optional arguments:
   --skip-build          Skip the image build step.
   --force-build         Enforce the image build step, regardless of if the
                         image already is available either locally or remotely.
+  --list-images         print list of images to stdout. Images will not be
+                        built.
```
consideRatio, line 129: This is not a feature part of this PR, just updated documentation in the README.

chartpress.py
```diff
@@ -537,7 +536,7 @@ def build_chart(name, version=None, paths=None, long=False):
     return version
 
 
-def publish_pages(chart_name, chart_version, chart_repo_github_path, chart_repo_url, extra_message=''):
+def publish_pages(chart_name, chart_version, chart_repo_github_path, chart_repo_url, extra_message, force):
```
betatim, line 539:
```suggestion
def publish_pages(chart_name, chart_version, chart_repo_github_path, chart_repo_url, extra_message='', force=False):
```
would keep the function signature the same. I don't know who/what code relies on `extra_message` being an optional keyword argument. Instead of finding out it would be easier to not change the signature.

chartpress.py
```diff
@@ -566,20 +567,38 @@ def publish_pages(chart_name, chart_version, chart_repo_github_path, chart_repo_
...
-            'git', 'clone', '--no-checkout',
-            git_remote(chart_repo_github_path),
-            checkout_dir,
-        ],
-        # warning: if echoed, this call could reveal the github token
-        echo=True,
-    )
+    # clone/fetch the Helm chart repo and checkout its gh-pages branch, note the
+    # use of cwd (current working directory)
+    checkout_dir = f'{chart_name}-{chart_version}'
```
betatim, line 572: With all the changes from `some_string.format()` to f-strings: does chartpress specify a minimum Python version it requires? If yes we need to bump that up if it is too old.

chartpress.py
```diff
@@ -566,20 +567,38 @@ def publish_pages(chart_name, chart_version, chart_repo_github_path, chart_repo_
...
+    if not os.path.isdir(checkout_dir):
+        check_call(
+            [
+                'git', 'clone', '--no-checkout',
+                git_remote(chart_repo_github_path),
+                checkout_dir,
+            ],
+            # warning: if echoed, this call could reveal the github token
+            echo=True,
```
betatim, line 581: The comment makes me think that I don't want this call to be echoed, but the code is explicitly set to echo. Which one is correct?

chartpress.py
```diff
@@ -566,20 +567,38 @@ def publish_pages(chart_name, chart_version, chart_repo_github_path, chart_repo_
...
+        )
+    else:
+        check_call(['git', 'fetch'], cwd=checkout_dir, echo=True)
     check_call(['git', 'checkout', 'gh-pages'], cwd=checkout_dir, echo=True)
 
+    # check if there is any already published chart with chart_name
```
betatim, line 587:
```suggestion
    # check if a chart with the same name and version has already been published. If
    # there is, the behaviour depends on `-force-publish-chart`
```

tests/test_repo_interactions.py
```diff
@@ -127,8 +127,23 @@ def test_chartpress_run(git_repo, capfd):
     git_repo.git.stash("pop")
 
 
+    # verify usage of --publish-chart when the chart version exists in the chart
+    # repo already
+    out = _capture_output(
```
betatim, line 132: For my curiosity: why do the tests use `_capture_output()` instead of [the `capfd` fixture](https://docs.pytest.org/en/latest/reference.html#std-fixture-capfd)?

betatim:
Can you add a test for the "with force" case as well?

chartpress.py
consideRatio: Yepp I checked this, its py36 already which is part of the test suite!

chartpress.py
consideRatio: This function is always called with all its arguments passed explicitly with defaults from the args parser, I felt by not providing another set of defaults here we would avoid confusion of having a duplicated sets of defaults where only one is actually used.

This would be a breaking change of chartpress as a Python library though where this function is called directly, but I don't know anyone doing that and we have made several library API changes before.

I can switch back to providing defaults, but I think we are inconsistent with this and it will cause a bit of confusion, I know for me it has been a bit confusing to have defaults that wasn't used and sometimes have been out of sync with a CLI like this.

chartpress.py
consideRatio: Yeah... I don't know. I don't remember why it could risk showing the GitHub token. I can invest time to investigate this once and for all.

tests/test_repo_interactions.py
consideRatio: @betatim I made these tests without being very knowledgeable about pytest, perhaps the capfd could have been used instead but I wasn't aware. I'm still not =/

The implementation of the function is printed below. It contains quite a bit of custom logic.

```python
def _capture_output(args, capfd, expect_output=False):
    """
    Calls chartpress given provided arguments and captures the output during the
    call.
    """
    # clear cache of in memory cached functions
    # this allows us to better mimic the chartpress CLI behavior
    chartpress.image_needs_building.cache_clear()
    chartpress.image_needs_pushing.cache_clear()
    chartpress.latest_tag_or_mod_commit.cache_clear()

    # first flush past captured output, then run chartpress, and finally read
    # and save all output that came of it
    _, _ = capfd.readouterr()
    chartpress.main(args)
    out, err = capfd.readouterr()
    if not expect_output:
        assert out == ""

    # since the output was captured, print it back out again for debugging
    # purposes if a test fails for example
    header = f'--- chartpress {" ".join(args)} ---'
    footer = "-" * len(header)
    print()
    print(header)
    print("out:")
    print(out)
    print("err:")
    print(err, file=sys.stderr)
    print(footer)
    return err
```

chartpress.py
consideRatio: Investigation complete. Apparently it relates to this.

```python
def git_remote(git_repo):
    """Return the URL for remote git repository.

    Depending on the system setup it returns ssh or https remote.
    """
    # if not matching something/something
    # such as a local directory ".", then
    # simply return this unmodified.
    if not re.match(r'^[^/]+/[^/]+$', git_repo):
        return git_repo

    github_token = os.getenv(GITHUB_TOKEN_KEY)
    if github_token:
        return f'https://{github_token}@github.com/{git_repo}'
    return f'git@github.com:{git_repo}'
```

If GITHUB_TOKEN is set with a CI system as a secret, it will probably be censored, but I can make it censored in chartpress as well, I'll make that happen.

consideRatio: I set it to false, it was a security flaw for anyone using GITHUB_TOKEN in a CI system that wouldn't censor it based on being set as a secure environment variable.

I added a note to make it clear what it was about.

chartpress.py
consideRatio: I added it back as it is a trivial un-important change.

consideRatio:
Thank you for your review @betatim ! :heart: :tada:

chartpress.py
consideRatio: I ensured it is censored and added a test to make sure the censoring works as well.

manics (commented):
My only thought is whether attempting to overwrite an existing version chart should return an error code instead of being silently ignored, in the same way that publishing the same version of a Pypi or Conda package will fail by default?

OTOH the current behaviour of always overwriting is different from those package managers, so maybe it doesn't matter.

consideRatio:
> My only thought is whether attempting to overwrite an existing version chart should return an error code instead of being silently ignored, in the same way that publishing the same version of a Pypi or Conda package will fail by default?

While it is printing a message about either overwriting when force is applied and skipping publication if force isn't applied, I agree it is almost silent.

I think I like it as it is, if we would start failing, it would be a very breaking change that can force people to default to `--force-publish-chart` to not have such failures, because it would be wrong to do `chartpress ... || true` or similar as that can fail for another reason that you don't want to ignore by `|| true`. So I figure if we would let it fail, we need yet another option.

I think skipping is the most sensible default, and it would be the most logical default matching how we do with `--push` / `--force-push` also, even though a `docker push` defaults to overwriting the image tag.

Yeah having considered it a bit now I think the current PR implementation reflects the desired behavior.

tests/test_repo_interactions.py
betatim: Aha! `_capture_output()` runs chartpress and captures the output using the normal pytest way. Because of the leading `_` I thought it was a pytest internal function we pulled in and used directly. Thanks for explaining!

PR #106, opened 2020-11-20
Strip v prefixes for Chart.yaml versions for Helm3 compliancy"

### LINKED ISSUE #88: Trim away a leading v character in a git tags when suitable
pangeo-data/helm-chart's latest git tag was v0.2.2, this gave us a v0.2.2-n###.h###### version that is an invalid SemVer 2 version because it leads with v, and this makes it an invalid Helm 3 version.

### LINKED ISSUE #89: What to do when tags are out of branch or is prefixed with a v?
## Tags on commits not in current branch
Chartpress picks the version it sets, like 0.9.0-n123.gabc1234, based on the latest commit _on the current branch_. This means that if someone makes a patch release 0.9.1 with a cherry picked commit, it will be out of branch and chartpress will keep counting from 0.9.0. This is probably the preferred behavior, but it is a bit sensitive and requires the user to be aware about it.

I wonder if we perhaps should provide a warning and documentation about this in the readme?

## Invalid helm versions
Helm3 requires the Chart.yaml version to not have a leading `v`, so if someone makes a tag like v0.9.0 chartpress would still set the Chart.yaml version field to v0.9.0 but that would be invalid. So, what should we do about this? I think the right call is to allow the `v` in the tag but make the version set in chart.yaml be with the `v` stripped while letting the docker images have the `v` part of the tag still.

I'd love some input on this change.

This part is being closed by #106.

consideRatio:
Note that this PR contain the commits of #105 as well, which should be merged first I think.

Closes #88 by stripping Chart.yaml version and notifying user about doing that, and also failing loudly if Chart.yaml still is invalid SemVer2 after that small fix attempt.
Closes #89 by updating the readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,19 +28,20 @@ information.
 1. `tag`: If not directly set by `--tag`, it will be inferred from most recent
    commit that is tagged in the _current branch_, or be set to 0.0.1 if no
    commit is tagged.
-   1. If the `--tag` flag starts with a `v`, it will be removed when setting the
-      Chart.yaml version specifically. An error will be raised if it still isn't
-      a valid SemVer2 version because Helm 3 requires it.
+   1. If the `tag` has a leading `v` but is otherwise a valid
+      [SemVer2](https://semver.org) version, it will be stripped from Chart.yaml
+      before its set as Helm 3 requires Helm chart versions to be SemVer2
+      compliant.
 1. The latest commit modifying content in a _relevant path_ since `tag`.
    1. `n`: The latest commit's distance to the tagged commit, described as 3 or
       more numbers, prefixed with n.
-   1. `h`: The latest commit's abbreviated SHA hash, which is typically 7 or a
-      few more characters, prefixed with h.
+   1. `h`: The latest commit's abbreviated hash. which is often 7-8 characters,
+      prefixed with h.
 1. If `tag` (like `0.10.0` or `0.10.0-beta.1`) contains a `-`, a `tag.n.h`
-   format will be used instead of a `tag-n.h` format.
-1. If `tag` (like `v1.0.0`) is prefixed with a `v`, it will be stripped.
+   format will be used instead of a `tag-n.h` format to be SemVer 2 compliant.
 1. If `--long` is specified or not. If `--long` is specified, tagged commits
-   will be written out with the `n.h` part appended to it.
+   will be written out with the `n.h` part appended to it, looking something
+   like `n000.gabcd123`
 
 ### Examples chart versions and image tags
 
--- a/README.md
+++ b/README.md
@@ -25,21 +25,22 @@ Chartpress can do the following with the help of some configuration.
 Chartpress will infer chart versions and image tags using a few key pieces of
 information.
 
-1. `tag`: The latest commit that is tagged on the current branch, or 0.0.1 if no
-   tag was found.
-2. The latest commit that influenced anything on a path within the git
-   repository that matters to the chart version or image tag. The paths that
-   matters is determined using the image build contexts and additional specified
-   paths.
-   1. `n`: The latest commits commit distance count since the tag, described as
-      3 or more numbers, prefixed with n.
-   2. `h`: The latest commits abbreviated SHA hash, which is typically 7
-      characters, prefixed with h.
-3. If `--long` is specified or not. When it is specified tagged commits will be
-   written out with `n000.h<hash>` appended to it.
-4. If `tag` contains a `-`, `tag.n.h` will be used, and if not, `tag-n.h` will
-   be used. There should be exactly one `-` in the final version specification
-   to become a valid SemVer2 version.
+1. `tag`: If not directly set by `--tag`, it will be inferred from most recent
+   commit that is tagged in the _current branch_, or be set to 0.0.1 if no
+   commit is tagged.
+   1. If the `--tag` flag starts with a `v`, it will be removed when setting the
+      Chart.yaml version specifically. An error will be raised if it still isn't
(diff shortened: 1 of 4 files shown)


chartpress.py
```diff
@@ -61,6 +61,24 @@ def _check_call(cmd, **kwargs):
...
+    semver2 = re.compile(r"^(?P<major>0|[1-9]\d*)\.(?P<minor>0|[1-9]\d*)\.(?P<patch>0|[1-9]\d*)(?:-(?P<prerelease>(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+(?P<buildmetadata>[0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$")
+
+    if semver2.fullmatch(version):
+        return version
+
+    if version.startswith("v") and semver2.fullmatch(version[1:]):
+        _log('Stripped a "v" from the chart version to become SemVer 2 compliant as required by Helm 3.')
+        return version[1:]
+
+    raise ValueError(f'The Chart.yaml "{version}" required by Helm 3 to be SemVer2 compliant!')
```
consideRatio, line 79:
```suggestion
    raise ValueError(f'The version in Chart.yaml "{version}" isn't SemVer2 compliant as required by Helm 3!')
```

README.md
```diff
@@ -25,21 +25,22 @@ Chartpress can do the following with the help of some configuration.
...
-   written out with `n000.h<hash>` appended to it.
-4. If `tag` contains a `-`, `tag.n.h` will be used, and if not, `tag-n.h` will
-   be used. There should be exactly one `-` in the final version specification
-   to become a valid SemVer2 version.
+1. `tag`: If not directly set by `--tag`, it will be inferred from most recent
+   commit that is tagged in the _current branch_, or be set to 0.0.1 if no
+   commit is tagged.
+   1. If the `--tag` flag starts with a `v`, it will be removed when setting the
+      Chart.yaml version specifically. An error will be raised if it still isn't
+      a valid SemVer2 version because Helm 3 requires it.
```
consideRatio, line 33:
```suggestion
   1. If the `tag` has a leading `v` but is otherwise a valid
      [SemVer2](https://semver.org) version, it will be stripped from Chart.yaml
      before its set as Helm 3 requires Helm chart versions to be SemVer2
      compliant.
```

README.md
```diff
@@ -25,21 +25,22 @@ Chartpress can do the following with the help of some configuration.
...
+1. The latest commit modifying content in a _relevant path_ since `tag`.
+   1. `n`: The latest commit's distance to the tagged commit, described as 3 or
+      more numbers, prefixed with n.
+   1. `h`: The latest commit's abbreviated SHA hash, which is typically 7 or a
+      few more characters, prefixed with h.
```
consideRatio, line 38:
```suggestion
   1. `h`: The latest commit's abbreviated hash. which is often 7-8 characters,
      prefixed with h.
```

README.md
```diff
@@ -25,21 +25,22 @@ Chartpress can do the following with the help of some configuration.
...
+1. If `tag` (like `0.10.0` or `0.10.0-beta.1`) contains a `-`, a `tag.n.h`
+   format will be used instead of a `tag-n.h` format.
```
consideRatio, line 40:
```suggestion
1. If `tag` (like `0.10.0` or `0.10.0-beta.1`) contains a `-`, a `tag.n.h`
   format will be used instead of a `tag-n.h` format to be SemVer 2 compliant.
```

README.md
```diff
@@ -25,21 +25,22 @@ Chartpress can do the following with the help of some configuration.
...
+1. If `tag` (like `v1.0.0`) is prefixed with a `v`, it will be stripped.
```
consideRatio, line 41:
```suggestion
```

README.md
```diff
@@ -25,21 +25,22 @@ Chartpress can do the following with the help of some configuration.
...
+1. If `--long` is specified or not. If `--long` is specified, tagged commits
+   will be written out with the `n.h` part appended to it.
```
consideRatio, line 43:
```suggestion
1. If `--long` is specified or not. If `--long` is specified, tagged commits
   will be written out with the `n.h` part appended to it, looking something
   like `n000.gabcd123`
```

chartpress.py
```diff
@@ -61,6 +61,24 @@ def _check_call(cmd, **kwargs):
...
+    semver2 = re.compile(r"^(?P<major>0|[1-9]\d*)\.(?P<minor>0|[1-9]\d*)\.(?P<patch>0|[1-9]\d*)(?:-(?P<prerelease>(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+(?P<buildmetadata>[0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$")
+
+    if semver2.fullmatch(version):
+        return version
+
+    if version.startswith("v") and semver2.fullmatch(version[1:]):
+        _log('Stripped a "v" from the chart version to become SemVer 2 compliant as required by Helm 3.')
+        return version[1:]
+
+    raise ValueError(f'The version in Chart.yaml "{version}" isn't SemVer2 compliant as required by Helm 3!')
```
consideRatio, line 79:
```suggestion
    raise ValueError(f"The version in Chart.yaml '{version}' isn't SemVer2 compliant as required by Helm 3!")
```

PR #108, opened 2020-12-03
fix: compute tag per image

danielnorberg:
I was trying to use `chartpress` to build binderhub with some (committed) changes to the builder, which I expected to trigger the building of a new `binderhub` image, but it did not. 

So I debugged `chartpress` and noticed that the `build_images` method would assign a value to `tag` using the first image in `images`. As this was the `cleaner` image, my builder changes were not included and as a result an older commit was assigned to `tag`. That same tag was then also used for the `binderhub` image and no new image built (as it already existed). 

In `jypterhub/binderhub` with a (committed) change to `builder.py`:

Before this PR:
```
$ chartpress
Updating binderhub/Chart.yaml: version: 0.2.0-n433.h38cd307
Skipping build for jupyterhub/k8s-image-cleaner:0.2.0-n415.hac2a96f, it already exists
Skipping build for jupyterhub/k8s-binderhub:0.2.0-n415.hac2a96f, it already exists
Updating binderhub/values.yaml: imageCleaner.image.tag: {'repository': 'jupyterhub/k8s-image-cleaner', 'tag': '0.2.0-n415.hac2a96f'}
Updating binderhub/values.yaml: image.tag: {'repository': 'jupyterhub/k8s-binderhub', 'tag': '0.2.0-n415.hac2a96f'}
```

Note that no image was built.

With this PR:
```
$ chartpress
Updating binderhub/Chart.yaml: version: 0.2.0-n433.h38cd307
Skipping build for jupyterhub/k8s-image-cleaner:0.2.0-n415.hac2a96f, it already exists
$> docker build -t jupyterhub/k8s-binderhub:0.2.0-n433.h38cd307 .. -f images/binderhub/Dockerfile
Sending build context to Docker daemon  9.695MB
Step 1/21 : ARG DIST=buster
Step 2/21 : FROM python:3.8-$DIST as build-stage
...
```

I.e., with this PR, chartpress builds a new binderhub image even if there's only been code changes that only affect the binderhub image.

### DIFF
--- a/chartpress.py
+++ b/chartpress.py
@@ -449,7 +449,9 @@ def build_images(prefix, images, tag=None, push=False, force_push=False, force_b
 
         # decide a tag string
         if tag is None:
-            tag = _get_identifier_from_paths(*all_image_paths, long=long)
+            image_tag = _get_identifier_from_paths(*all_image_paths, long=long)
+        else:
+            image_tag = tag
 
         image_name = options.get('imageName', prefix + name)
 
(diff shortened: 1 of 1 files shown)


minrk:
Good catch, thanks!

consideRatio:
@danielnorberg =D we made exactly the same changes to the code in parallel! This is already released as 1.0.1. See 8196119 vs 1c836db! Thanks for catching this also :)

danielnorberg:
@consideRatio Great! And impressive that we managed to make exactly the same code change! 🙌 😄

PR #109, opened 2020-12-04
only apply strict version checking for charts to be published

minrk:
and clarify that the issue is only about installing charts from repositories not deploying charts themselves

related to https://github.com/jupyterhub/mybinder.org-deploy/issues/1725


consideRatio:
Thank you @minrk

PR #110, opened 2020-12-07
allow standard `--no-build` prefix for disabling builds

minrk:
`--no-$feature` is a standard format for disabling something on a CLI, so allow this form in addition to the existing `--skip-build`.

I've typed `chartpress --no-build` *so many times* at this point, I thought I might as well make it work :)


consideRatio:
This LGTM - can you do do a README.md update of `chartpress --help` output also?

consideRatio:
Released and made available as part of 1.0.3!

PR #119, opened 2021-04-01
Add pre-commit

manics:
Configuration is a direct copy of https://github.com/jupyterhub/kubespawner/blob/98bb5773f0d1c8e0c065c7259511ea38103eabe9/.pre-commit-config.yaml

Is it worth it the overhead of pre-commit for just one source file? I don't know, but it's about the same amount of work to open an issue to discuss it as it is to open this PR even if we don't merge it :smile:

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -25,7 +25,14 @@ defaults:
     shell: bash
 
 jobs:
-  chart_tests:
+  pre-commit:
+    runs-on: ubuntu-20.04
+    steps:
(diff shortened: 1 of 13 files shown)


consideRatio:
I appreciate that we systematically use it pre-commit etc at this point!

consideRatio:
Thanks @manics! I two commits, one to run pre-commit in in a CI job, and one to remove irrelevant logic that was introduced as a mistake by me when i thought `sh` was the default shell.

PR #120, opened 2021-04-02
Add mock tests for public chartpress methods that call commands

manics:
Adds some unit tests that mock the `chartpress._check_call` method, which means we can test more combinations of input arguments without complicated setups.

Ideally these tests should test all variations of all input arguments, but we can add them in future whenever the method is modified (unless someone else wants to push to this PR).

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -6,6 +6,8 @@ from distutils.dir_util import copy_tree
 import git
 import pytest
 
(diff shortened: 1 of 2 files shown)


consideRatio (approved):
Aaaa this make use of some quite cool ideas! LGTM! Go for merge if you feel it's ready!

PR #123, opened 2021-04-04
Optionally use buildx to build for multiple platforms

manics:
- New argument `--builder` to switch between `docker build` and `docker buildx`. I've used a `choice` instead of a flag in case we want to add more builders in future.
- New argument `--platform` that takes platforms accepted by `docker buildx --platform`


consideRatio:
The suggested API makes sense to me!

chartpress.py
```diff
@@ -293,14 +305,41 @@ def build_image(image_spec, context_path, dockerfile_path=None, build_args=None)
...
+    if platforms:
+        cmd.extend(["--platform", ",".join(platforms)])
+        # Limitations of docker buildx 0.5.1:
+        # - Can't load into the local Docker host and push to a registry at the
+        #   same time
+        # - Can't load multiple platforms into the local docker host
+        if push:
+            cmd.append("--push")
+        elif len(platforms) == 1:
+            cmd.append("--load")
```
consideRatio, line 337: I love the comments!!! Good you included the version number as well when this was the case.

chartpress.py
```diff
@@ -458,6 +497,9 @@ def build_images(
     force_build=False,
     skip_build=False,
     long=False,
+    *,
+    builder=BUILDERS[0],
```
consideRatio, line 501: You made an if statement referencing "docker-build" the string öiteral above, but here you reference the constant. A bit inconsistent, perhaps go string literal here also for readability?

chartpress.py
```diff
@@ -519,8 +566,10 @@ def build_images(
 
         image_spec = f"{image_name}:{image_tag}"
 
-        # build image
-        if force_build or _image_needs_building(image_spec):
+        # build image and optionally push image
+        # Since docker buildx builds for multiple platforms we can't tell whether the
+        # image already exists in the host so always build and rely on caching
+        if builder != BUILDERS[0] or force_build or _image_needs_building(image_spec):
```
consideRatio, line 572: Perhaps the function image needs building should contain this new logic?

consideRatio (commented):
Nice this looks good overall so far in my mind, i left some comments about thoughts that popped up.

manics:
I've switched to an `Enum` which I think is a lot clearer! Also added some tests.

I think this is ready to review now. I'll add an option to skip an image (for the Z2JH singleuser image) in a separate PR.

consideRatio (commented):
This looks good to me except that I didn't understand a part about providing `*` in a list of parameters for a few functions.

I guess there will be a sensible explanation for that though, and then this LGTM!

chartpress.py
```diff
@@ -272,7 +282,16 @@ def _get_all_chart_paths(options):
     return list(set(paths))
 
 
-def build_image(image_spec, context_path, dockerfile_path=None, build_args=None):
+def build_image(
+    image_spec,
+    context_path,
+    dockerfile_path=None,
+    build_args=None,
+    *,
```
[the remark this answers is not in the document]
manics: It means all arguments after the `*` are keyword _only_ arguments: https://python-3-for-scientists.readthedocs.io/en/latest/python3_advanced.html (or https://www.python.org/dev/peps/pep-3102/ if you really want the hard details :smile: ).

Advantages of using it for functions with long argument lists include:
- no chance of accidentally calling the function with positional arguments in the wrong order, you _must_ pass them as keywords
- easier to refactor the function signature in future, if you want to insert a new argument or reorder the existing ones for legibility you can do so without a breaking change since the position of the arguments doesn't matter.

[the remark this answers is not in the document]
consideRatio: Ah I like it, thanks for helping me learn about this!

PR #124, opened 2021-04-10
Add skipPlatforms option for multi-platform docker buildx images

manics:
In Z2JH the [singleuser image can only be built for `linux/amd64`](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/08faa6f197e478ca28ce23d083287d37096cbb36/images/singleuser-sample/Dockerfile#L1) as it depends on https://github.com/jupyter/docker-stacks/. This means we need a way to omit selected images when building a multi-image multi-architecture chart.

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,11 +86,18 @@ In a directory containing a `chartpress.yaml`, run:
 to build your chart(s) and image(s). Add `--push` to publish images to docker
 hub and `--publish-chart` to publish the chart and index to gh-pages.
 
+<!--
+To update this help output run
+COLUMNS=80 chartpress --help
+-->
+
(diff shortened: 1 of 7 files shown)


consideRatio:
Will review this more in depth soon but quickly wanted to say i agree with the configuration option idea.

consideRatio (approved):
I love being able to learn new things and techniques etc by reviewing your PRs :)

This looks great to me! I'll go for a merge.

Do you consider it to be time for for a release at this point?

manics:
I think so!
https://github.com/jupyterhub/chartpress/pull/125

PR #127, opened 2021-06-22
Define pre-commit hook

minrk:
Copies over pre-commit hook from [zero-to-jupyterhub](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/c68e9a4db6ca54c464cfd67ec7a5a0c63d47f874/.pre-commit-config.yaml#L42-L50) so it can be re-used.

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -3,5 +3,5 @@
   files: (Chart|chartpress|values).ya?ml
   description: Run `chartpress --reset` to clean up helm charts before committing.
   entry: chartpress --reset
-  language: python_venv
+  language: python
   pass_filenames: false
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,0 +1,7 @@
+- id: chartpress
+  name: chartpress --reset
+  files: (Chart|chartpress|values).ya?ml
+  description: Run `chartpress --reset` to clean up helm charts before committing.
+  entry: chartpress --reset
+  language: python_venv
+  pass_filenames: false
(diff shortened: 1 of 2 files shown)


.pre-commit-hooks.yaml
```diff
@@ -0,0 +1,7 @@
+- id: chartpress
+  name: chartpress --reset
+  files: (Chart|chartpress|values).ya?ml
+  description: Run `chartpress --reset` to clean up helm charts before committing.
+  entry: chartpress --reset
```
consideRatio, line 5: I considered this but figured it wouldnt work well as chartpress must be executed next to the chartprwss.yaml file, but binserhub and other git repos doesnt have it in the root folder.

Perhaps a param will be required for where one chartpress.yaml file is found.

consideRatio:
I understand from [this pre-commit documentation](https://pre-commit.com/#python) that we kind of need to have this hook defined here to get chartpress automatically installed by pre-commit.ci service.

.pre-commit-hooks.yaml
```diff
@@ -0,0 +1,7 @@
...
+  language: python_venv
```
consideRatio, line 6:
```suggestion
  language: python
```

Apparently `python_venv` is deprecated and an alias of `python` - see https://pre-commit.com/#python_venv

.pre-commit-hooks.yaml
```diff
@@ -0,0 +1,7 @@
+- id: chartpress
```
consideRatio, line 1:
```suggestion
# This is a pre-commit hook definition, for more details see:
# https://pre-commit.com/#new-hooks
#
# The purpose of this hook is to restore Chart.yaml and values.yaml files
# that may have been updated by chartpress.
#
- id: chartpress
```

PR #136, opened 2021-07-24
_image_needs_pushing: check platforms when using docker buildx

### LINKED ISSUE #132: Avoid rebuilding an image already available in registry when two --platform flags are passed
### Bug description (Broken expectation)

Chartpress doesn't recognize if an image is already available when passing multiple platforms and will instead rebuild the image.

### How to reproduce

- Make sure to update the imagePrefix below to a location you have push rights
- Make sure to install buildx and configure so you can use arm64

```
mkdir -p /tmp/test-chartpress/images/test-image
cd /tmp/test-chartpress

helm create test-chart

cat <<EOF > chartpress.yaml
charts:
  - name: test-chart
    imagePrefix: quay.io/consideratio/
    images:
      test-image: {}
EOF

cat <<EOF > images/test-image/Dockerfile
FROM python:3.8-slim
EOF

git init
git add .
git commit -m "initial commit"

# First push someplace
chartpress --builder docker-buildx --platform linux/amd64 --platform linux/arm64 --push

# I'd like this to not even try to build, cache or not, as the
# image is now available in the registry since the last command,
# which is what i want to be detected or similar at least unless
# we pass --force-push where that detection is ignored
chartpress --builder docker-buildx --platform linux/amd64 --platform linux/arm64 --push
```

### Relevant code

https://github.com/jupyterhub/chartpress/blob/cf43a15031d2703f7727133faeec5b46e3daab95/chartpress.py#L379-L382

https://github.com/jupyterhub/chartpress/blob/cf43a15031d2703f7727133faeec5b46e3daab95/chartpress.py#L412-L415

### Suggested change

I think we should opt out of this better safe than sorry logic and instead use a default behavior with higher performance and potentially a warning saying that it may have made the wrong decision, where the user can fall back to `--force-build` and `--force-push` if wanted.

@manics, you have become very knowledgeable about the multi-arch builds. WDYT?

manics:
Closes #132

Not yet tested....

### DIFF
--- a/chartpress.py
+++ b/chartpress.py
@@ -367,8 +367,15 @@ def _get_docker_client():
     return docker.from_env()
 
 
-@lru_cache()
 def _image_needs_pushing(image, platforms):
+    # @lru_cache won't allow a list as a function parameter
+    if platforms:
+        platforms = frozenset(platforms)
+    return _image_needs_pushing_cached(image, platforms)
+
+
+@lru_cache()
+def _image_needs_pushing_cached(image, platforms):
     """
     Returns a boolean whether an image needs pushing by checking if the image
     exists on the image registry.
@@ -399,7 +406,6 @@ def _image_needs_pushing(image, platforms):
         return False
 
 
-@lru_cache()
 def _image_needs_building(image, platforms):
     """
     Returns a boolean whether an image needs building by checking if the image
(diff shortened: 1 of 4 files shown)


consideRatio:
Since we want to work with a frozen set rather than a l list, should we perhaps instead convert the platforms list to a frozen set directly after `args = argparser.parse_args(args)` and adjust various functions relating to `platform`?

consideRatio:
@manics this looks good, and it successfully opted to not push when i tried out the reproduction steps in https://github.com/jupyterhub/chartpress/issues/132#issue-951247360 (that i had small errors in, but fixed them just now).

If you think its worth the time and effort, and it is a sensible idea to refactor platforms to be a frozenset before the helper functions checking if build/push needs to be done are called, then do that but otherwise I think we should go for a merge!

manics:
Good point, done!

consideRatio:
@manics wieeee!!! Thanks for your work on multi platform stuff, I love it!!

Issue #141, opened 2021-12-31
Ability to include extra options to `docker build`


adamblake:
### Proposed change

I would like a way to include other docker commands when building a particular image. The reason I want this is because I access a private repo when building the image and I want to [forward an SSH connection to the builder](https://docs.docker.com/develop/develop-images/build_enhancements/#using-ssh-to-access-private-data-in-builds).

I am imagining that this could be accomplished in chartpress.yaml with an `extraOptions` key that allows passing other `docker build` options by using the option name as the key and the value as the value. Here is an example that would add `--ssh default` to the build command:

```yaml
charts:
  - name: chart-name
    images:
      image-name:
        extraOptions:
          ssh: default
```

### Alternative options

There are two other ways that I thought this could work:

**Dedicated keys**:
You could add each `docker build` option as an optional key, but then you have to keep up with Docker as it adds to or changes the API. This would be convenient for some users but not many, and not convenient to the maintainers.

**Arbitrary code**:
Instead of specifying each extra option and its key and value, you could just have `extraOptions` take a string that gets appended to the build command. I think this results in a much less readable approach when more options are added. In fact, I don't really even like it for the single option here:

```yaml
charts:
  - name: chart-name
    images:
      image-name:
        extraOptions: "--ssh default"
```

### Who would use this feature?

Anyone who would like to use more of the functionality of `docker build`, especially those looking to use `--secrets` and `--ssh` when building their images.

### (Optional): Suggest a solution

I think this is actually a fairly straightforward implementation. I am happy to write it and submit a pull request if you like the approach. Essentially I would modify `build_image` like this:

```python
def build_image(
    image_spec,
    context_path,
    dockerfile_path=None,
    build_args=None,
    extra_options=None,
    *,
    push=False,
    builder=Builder.DOCKER_BUILD,
    platforms=None,
):
    
    ...
    
    for k, v in (extra_options or {}).items():
        cmd += [f"--{k}", f"{v}"]

    ...
```

And then modify the calls to `build_image()` to get the `extraOptions` from *chartpress.yaml*. 

#### Caveats to this solution

Some build arguments are only applicable to more recent versions of Docker, e.g. `--ssh` came with Docker 18.09. I think this should be left to the user to understand.

PR #142, opened 2021-12-31
Allow extra options to be passed to docker build

adamblake:
This is a first pass at implementing #141 (Ability to include extra options to `docker build`). It allows users to specify extra build options in *chartpress.yaml* on a per-image basis. This is useful for taking advantage of other build options that chartpress does not explicitly support. 

I am not very familiar with pytest, so there are no tests included here. I can confirm that it is working using the --ssh option to clone a private GitHub repo in my local builds.

### DIFF
--- a/README.md
+++ b/README.md
@@ -200,12 +200,6 @@ charts:
           - --label=maintainer=octocat
           - --label=ref={TAG}-{LAST_COMMIT}
           - --rm
-          # above is equivalent to
-          - --label
-          - maintainer=octocat
-          - --label
-          - ref={TAG}-{LAST_COMMIT}
-          - --rm
         # contextPath is the path to the directory that is to be considered the
         # current working directory during the build process of the Dockerfile.
         # This is by default the folder of the Dockerfile. This path should be
--- a/README.md
+++ b/README.md
@@ -192,13 +192,20 @@ charts:
         buildArgs:
           MY_STATIC_BUILD_ARG: "hello world"
           MY_DYNAMIC_BUILD_ARG: "{TAG}-{LAST_COMMIT}"
-        # Build options to be passed to docker build as --<key> <value>. This
-        # allows you to use all available docker options for your build.
-        # The example below adds "--ssh default=secrets/private-key" to the build 
-        # command. Note that for flags like "--ssh" which take a key-value pair,
-        # you must use the "key=value" syntax and not "key value".
-        extraOptions:
-          ssh: default=secrets/private-key
+        # Build options to be passed to the docker build command. Pass a list
+        # of strings to be appended to the end of the build command. These are
+        # passed directly to the command line, so prepend each option with "--"
+        # like in the examples below. TAG and LAST_COMMIT are expandable.
+        extraBuildCommandOptions:
+          - --label=maintainer=octocat
+          - --label=ref={TAG}-{LAST_COMMIT}
+          - --rm
+          # above is equivalent to
+          - --label
+          - maintainer=octocat
+          - --label
+          - ref={TAG}-{LAST_COMMIT}
+          - --rm
         # contextPath is the path to the directory that is to be considered the
         # current working directory during the build process of the Dockerfile.
         # This is by default the folder of the Dockerfile. This path should be
--- a/README.md
+++ b/README.md
@@ -192,6 +192,13 @@ charts:
         buildArgs:
           MY_STATIC_BUILD_ARG: "hello world"
           MY_DYNAMIC_BUILD_ARG: "{TAG}-{LAST_COMMIT}"
+        # Build options to be passed to docker build as --<key> <value>. This
+        # allows you to use all available docker options for your build.
+        # The example below adds "--ssh default=secrets/private-key" to the build 
+        # command. Note that for flags like "--ssh" which take a key-value pair,
+        # you must use the "key=value" syntax and not "key value".
+        extraOptions:
+          ssh: default=secrets/private-key
         # contextPath is the path to the directory that is to be considered the
         # current working directory during the build process of the Dockerfile.
         # This is by default the folder of the Dockerfile. This path should be
--- a/chartpress.py
+++ b/chartpress.py
@@ -240,14 +240,13 @@ def _get_image_extra_build_command_options(image_options, ns):
     """
     Render extraBuildCommandOptions from chartpress.yaml that could be
     templates, using the provided namespace that contains keys with dynamic
-    values such as LAST_COMMIT or TAG. The rendered options will be converted
-    to a list of strings that can be added to the build command list.
+    values such as LAST_COMMIT or TAG.
 
     Args:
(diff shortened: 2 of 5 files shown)


manics:
A disadvantage of using a dict is that repeated arguments aren't allowed.

consideRatio:
I support adding this feature!

Good point @manics! Go for a list, and the name should be more explicit i think. Args is too vague as it there are both docker build args and docker push args, and even helm cli args of possible relevance.

Perhaps dockerBuildArgs or maybe with an extra prefix so it is extraDockerBuildArgs.

adamblake:
@consideRatio I chose `extraOptions` to intentionally separate the idea of "build arguments" (i.e. `--build-arg=`) and "`docker build` options" (i.e. arbitrary options). Calling it `extraBuildArgs` would make it easy to confuse with the other *chartpressl.yaml* key `buildArgs` which references the build arguments and not the options.

Since the docker build documentation calls these "options" I thought it was a good choice (here's the command interface from [their docs](https://docs.docker.com/engine/reference/commandline/build/) `docker build [OPTIONS] PATH | URL | -`). I also didn't think that "docker" needed to be added to the key because it's nested in the `images` key and the other related keys don't have it (e.g. `buildArgs`). 

Here's a more full example of a *chartpress.yaml* that would use this key. I made the `extraOptions` a list like @manics suggested. I also considered that some options have default values so the interface should allow you to pass no value to them.

```yaml
charts:
  - name: some-jupyter-hub
    imagePrefix: a-nice-prefix/
    images:
      k8s-hub:
        valuesPath: hub.image
        buildArgs:
          TIMEOUT: "15m"
          COMMIT_REF: "1.4.5"
        extraOptions:
          - rm  # uses default value
          - label: "maintainer=octocat"
          - label: "version=0.1.0"
          - ssh: "default=secrets/secret-auth-key"    
```

I'll implement the list, and please let me know what you think would be the best key name!

consideRatio:
Great point about the `--build-args` flag @adamblake! Hmmm, I'd love for it to be quite self-explanatory, but given that we also have `--build-args` causing some confusion I figure we would need to have it be `extraDockerBuildOptions`.

@manics, what do you think about the naming? I'm generally in favor of any name, almost no matter how long it is, that makes it quite clear directly from the name what it does.

I think I'd prefer an API that minimizes complexity and embedded logic, so for example...

```yaml
charts:
  - name: some-jupyter-hub
    imagePrefix: a-nice-prefix/
    images:
      k8s-hub:
        valuesPath: hub.image
        buildArgs:
          TIMEOUT: "15m"
          COMMIT_REF: "1.4.5"
        extraBuildCommandArgs:
          - rm
          - label=maintainer=octocat
          - label=version=0.1.0
          - ssh=default=secrets/secret-auth-key
# Equivalently, one can split apart a args key/value into
# two separate elements in the list.
#      extraBuildCommandArgs:
#        - rm
#        - label
#        - maintainer=octocat
#        - label
#        - version=0.1.0
#        - ssh
#        - default=secrets/secret-auth-key
```

We currently have buildArgs support a structure like above though, where it's key/values are the value of the --build-arg flag... hmmm....

Brainstorming:
- extraOptions | extraArgs
- extraBuildOptions | extraBuildArgs
- extraDockerBuildOptions | extraDockerBuildArgs
- extraBuildCommandOptions | extraBuildCommandArgs
- extraDockerBuildCommandOptions | extraDockerBuildCommandArgs

Perhaps @manics @minrk could join in and opine on this?

1. We need a config name
2. We need to choose to accept:
   - strings only
   - dictionaries with a single key (null value -> empty string value)
   - strings or a dictionary with a single non-null key

I'm leaning towards `extraDockerBuildCommandArgs` and supporting strings only, where I'm quite clearly opined towards strings only but less clearly opined about the config name.

manics:
I think `extraDockerBuildCommandArgs` / `extraDockerBuildsArgs` / `extraDockerArgs` / something along those lines is fine, I'm happy to trust @consideRatio's judgement on the final choice.

I think the mix of single values and dicts adds unnecessary complexity. These arguments are being passed directly to a command line program which expects a list of strings, so let's keep it as that?

minrk:
I do think a simple opaque list of strings to pass through is the way to go.

One more option is to specify `dockerBuildCommand` itself the full CLI list (`['docker', 'buildx', 'build', '--whatever-other-flags']`). Then we don't have to decide whether to call them Args or Options or Flags :). For confusion with `buildArgs` which I think is unambiguous and has the right name This would allow things like specifying the abspath to docker, etc. if that matters.

adamblake:
It seems like a list of strings and a name like `extraBuildCommandOptions` is the consensus. I'll implement that and add some tests.

(I'll leave the other suggestion about the full command for now because I'm not sure how that would play with the other chartpress keys.)

consideRatio:
> It seems like a list of strings and a name like extraBuildCommandOptions is the consensus. I'll implement that and add some tests.

Sounds great to me, thank you @adamblake!

I think the name `extraBuildCommandOptions` is okay as it manages to a) avoid the confusion about `--build-args`, b) convey its about the `docker build` part, c) convey its in addition to other flags.

adamblake:
I would fix the issue with the pre_commit GitHub Action, but I don't understand the error. I did run `pre-commit run -a` before committing though, and there were no errors.

chartpress.py
```diff
@@ -322,11 +340,10 @@ def build_image(
...
         Dictionary of docker build arguments.
-    extra_options (dict, optional):
-        Dictionary of other docker build options to use. Each key should be a
-        valid option to docker build, like "ssh" for the "--ssh" option. For 
-        flags like "--ssh" which take a key-value pair, make sure to use the
-        "key=value" syntax and not the "key value" syntax.
+    extra_build_command_options (list, optional):
+        List of other docker build options to use. Each item can be either a
+        string with the option name (e.g. "rm") or a dictionary with the key
+        and value (e.g. "label: 'maintainer=octocat'").
```
consideRatio, line 346: A dict is still mentioned here

adamblake: Thanks -- I looked for similar errors in the docstrings and fixed those as well.

README.md
```diff
@@ -192,6 +192,20 @@ charts:
...
+        extraBuildCommandOptions:
+          - --label=maintainer=octocat
+          - --label=ref={TAG}-{LAST_COMMIT}
+          - --rm
+          # above is equivalent to
+          - --label
+          - maintainer=octocat
+          - --label
+          - ref={TAG}-{LAST_COMMIT}
+          - --rm
```
consideRatio, line 208: Now that we have pure string appending without also prefixing the strings with `--`, this is easy to understand in general I think, so I'll remove this part that now feels redundant.

```suggestion
```

consideRatio (approved):

consideRatio:
Thank you @adamblake for the PR and @manics and @minrk for review!! :tada: :heart:

Issue #143, opened 2022-02-21
Chartpress only considers tags in the current branch - it's too narrow, right?


consideRatio:
When chartpress considers git tags to get a version for the images/helm chart, I think the logic it use is to take the latest tag on the current branch. I think we need to assume that we should use the latest tag on any branch instead.

https://github.com/jupyterhub/chartpress/blob/9eb6b0a631df439d929d4b5651794f0a1d6a1a25/chartpress.py#L150-L157

minrk:
> I think we need to assume that we should use the latest tag on any branch instead.


I'm not so sure. Let's say we have a `1.x` backport branch, which would be created from `1.0` or later (whichever was the last commit in master before forking to start the backport). For example:

<img width="149" alt="Screen Shot 2022-02-21 at 12 38 16" src="https://user-images.githubusercontent.com/151929/154948438-a537f9a3-aa05-43fc-a1c3-886a49f63a92.png">

where there's a 1.x and main branch. The 1.x branch has `1.1.1` and `1.2.1` tags. Both branches share a `1.0.0` tag, and the main branch only has `2.0.0`.

The 'latest' tag in time is 1.1.2. This tag is only on the 1.x branch. The latest _version_ tag is 2.0.0. This is only on main. Logically, the `1.x` branch would get `1.1.2` as the latest tag, while `main` would get `2.0.0`. This is what would happen with the current implementation using `git describe`. If we used _either_ the latest-in-time (1.1.2) or latest-in-version (2.0.0) on both branches, one would always be wrong.

The case where we could arguably get the wrong answer is _before_ the first tag on main is published, but after a backport release, so only `1.1.1` is published, no `2.0.0`:

<img width="152" alt="Screen Shot 2022-02-21 at 12 46 22" src="https://user-images.githubusercontent.com/151929/154949369-50e9c182-157d-4293-9e77-9a19ddac5dd9.png">

Here, `1.x` will get everything right, while `main` would still get `1.0.0`. Arguably, we could expect `1.1.1` on main, as that's the most recent release. But that doesn't make perfect sense, because main is _not_ a descendent of 1.1.1 on the graph. We will get the _accurate_ information that this is 5 commits after 1.0.0, but not any information about how it relates to 1.1.1, and, perhaps most importantly, the prerelease versions won't sort correctly until there's been a new tag on main.

[...]

PR #145, opened 2022-02-21
Add tests with dev tags and backport branches

minrk:
everything seems to work without fixing, but good to test anyway

related to #143, looks like we handle `2.0.0-dev` tags just fine.


minrk:
needs #144 for the gh-pages test to pass

PR #150, opened 2022-03-01
Add `useChartVersion` and change appended version suffix (now like `1.2.3-0.dev.git.3.h123`)

### LINKED ISSUE #174: Reimplement current `chartpress --reset --tag`
From https://github.com/jupyterhub/chartpress/pull/150#issuecomment-1059150383 in #150 @minrk and I discussed `chartpress --reset --tag`, which currently updates the Chart.yaml version, but not the images.

> > Only the Helm chart's version needs to be updated, not the images.
>
> True, and this is what chartpress --reset --tag x currently does (chart.resetVersion and chart.resetTag are set via different paths). This might be a bug! But we can make a new entrypoint, or add further clarification of what to do when, if that would be helpful.

I think its hard to understand and document the current `chartpress --reset` interaction with `--tag`, and it feels buggy that it influences the Chart.yaml version but not the images tag. So maybe we can iterate on this?

What about providing something like this, where we have  a use ase for `--reset-chart-version` specifically with #150?

```
chartpress --reset --reset-chart-version=1.2.3 --reset-image-tag=1.2.3
```

minrk:
Instead of `1.2.3-n003.habc` produce `1.2.3-0.dev.git.3.habc`

## Problem

Our current dev versions don't sort correctly for various reasons:

- `n999` and `n1000` sort alphabetically, resulting in incorrect ordering
- prereleases are relative to the _most recent_ tag, which results in incorrect ordering after every release (`1.2.3` > `1.2.3-n003.habc`, when it should be the other way around) unless repos publish a prerelease tag immediately after every stable release

The only way to get prereleases to sort correctly is to produce them relative to the _next_ release, not the _previous_ release, which we get from `git describe`.

Plus, the version information is misleading - there's no indication of whether a dev version is preparing for a minor release or a major one - only that it's _after_ some version in the past.

## Proposal

- Use numerical field instead of `n012` zero-padded text fields for commit count. New tag scheme appends `.git.{n}.h{sha}` to prereleases (e.g. `1.2.3-alpha.1` -> `1.2.3-alpha.1.git.5.habc`), so commit counts are compared numerically instead of alphabetically.
- Uses leading `0.dev.` for base prerelease versions, if building a dev version from a non-prerelease version (e.g. `1.2.3` -> `1.2.3-0.dev.git.5.habc`) so that it sorts before `alpha`
- new config (opt-in) `useChartVersion` to use the version in Chart.yaml as the base version, instead of the most recent git tag.
- When using the new mode, `chartpress --reset` resets the version by stripping the `.git.{n}.h{sha}` suffix, leaving the base version alone, rather than clobbering with invalid `set-by-chartpress` tag.

Notably, this does _not_ solve the prerelease < stable release ordering by default, it only _allows_ repos to solve it by opting in to the `useChartVersion` config, and specifying the dev version in Chart.yaml, as is standard practice in software packages (1. tag a release, 2. set version to `$next-dev`). I think this is the right approach, becuse the next version is not guessable, so it has to come from the maintaners somehow, either:

1. in Chart.yaml (this implementation)
2. chartpress.yaml (also works, but Chart.yaml seems simpler to me)
3. in a git tag (forces unnatural release process)

**EDIT: original description**

When we added the `n00` prefix, I misunderstood semver restrictions about leading 0s. We can have leading 0s on fields that aren't _just_ digits (e.g. `0git` is okay, but `05` is not), so the prefix was only strictly necessary on the hash, not the count. And we can have fields that are exactly one 0 (`.0.` is okay, `.00` is not). So I think the `n00` prefix was not the best solution - `n.5` is better than `n005`. It's correctly sortable (`n.1000` > `n.999`, unlike `n1000` and `n999`), and separates the numerical field from the 'channel' field.

Further, once we have two fields, I think we should sort before 'alpha', so that a pre-alpha dev release (1.2.3-$tag) will sort before the following alpha (1.2.3-alpha.$tag). I chose `0git` for this, but I'm not sure that's best. Coud use `0.dev.git`

To this end, I propose changing our tag scheme from `n{n:03}.h{hash}` to `0git.{n}.h{hash}`:

- allows fully numerical fields to sort properly
- still avoids leading 0s on numerical fields
- prefix with `0git.` to sort before 'alpha.'

Other options:

1. No `0git.` prefix, just use numbers. This _might_ be confusing for a version that already has a number, like `1.0.0-alpha.1.5.habc` or `--long` tags with no commits like `1.0.0-0.habc`.
4. use just `0`: `1.0.0-alpha.1.0.5.habc`
5. use just `git` or `n` - I like these better aesthetically, but they have the problem that they sort after `alpha` and `beta`, so `1.2.3-git.5.habc > 1.2.3-alpha.1`

closes #174

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,22 +25,25 @@ Chartpress will infer chart versions and image tags using a few key pieces of
 information.
 
 1. `tag`: If not directly set by `--tag`, it will be inferred from most recent
-   commit that is tagged in the _current branch_, or be set to 0.0.1 if no
-   commit is tagged.
+   commit that is tagged in the _current branch_
+   (as determined by `git describe`)
+   or be set to 0.0.1 if no commit is tagged.
    1. If the `tag` has a leading `v` but is otherwise a valid
-      [SemVer2](https://semver.org) version, it will be stripped from Chart.yaml
+      [SemVer2](https://semver.org) version, the `v` will be stripped from Chart.yaml
       before its set as Helm 3 requires Helm chart versions to be SemVer2
       compliant.
 1. The latest commit modifying content in a _relevant path_ since `tag`.
-   1. `n`: The latest commit's distance to the tagged commit, described as 3 or
-      more numbers, prefixed with n.
+   1. `n`: The number of commits since the tagged commit, as an integer.
    1. `h`: The latest commit's abbreviated hash. which is often 7-8 characters,
-      prefixed with h.
-1. If `tag` (like `0.10.0` or `0.10.0-beta.1`) contains a `-`, a `tag.n.h`
-   format will be used instead of a `tag-n.h` format to be SemVer 2 compliant.
+      prefixed with `h`.
+1. If `tag` (like `0.10.0` or `0.10.0-beta.1`) contains `-`,
+   indicating that it is a prerelease,
+   the chartpress suffix will be added to the prerelease fields,
+   e.g. `0.10.0-beta.1.0git.5.habc123`
+   format will be used instead of a `tag-0git.n.h` format to be SemVer 2 compliant.
 1. If `--long` is specified or not. If `--long` is specified, tagged commits
-   will be written out with the `n.h` part appended to it, looking something
-   like `n000.gabcd123`
+   will always be written out with the `n.h` part appended to it, looking something
+   like `1.0.0-0git.0.habcd123`
 
 ### Examples chart versions and image tags
 
@@ -49,13 +52,13 @@ order that could come from using chartpress.
 
 ```
 0.8.0
-0.8.0-n004.hasdf123
-0.8.0-n010.hsdfg234
+0.8.0-0git.4.hasdf123
+0.8.0-0git.10.hsdfg234
 0.9.0-beta.1
-0.9.0-beta.1.n001.hdfgh345
-0.9.0-beta.1.n005.hfghj456
+0.9.0-beta.1.0git.1.hdfgh345
+0.9.0-beta.1.0git.5.hfghj456
 0.9.0-beta.2
-0.9.0-beta.2.n001.hghjk567
+0.9.0-beta.2.0git.1.hghjk567
 0.9.0-beta.3
 0.9.0
 ```
--- a/README.md
+++ b/README.md
@@ -39,11 +39,16 @@ information.
 1. If `tag` (like `0.10.0` or `0.10.0-beta.1`) contains `-`,
    indicating that it is a prerelease,
    the chartpress suffix will be added to the prerelease fields,
-   e.g. `0.10.0-beta.1.0git.5.habc123`
-   format will be used instead of a `tag-0git.n.h` format to be SemVer 2 compliant.
+   e.g. `0.10.0-beta.1.git.5.habc123`
+   format will be used instead of a `tag-0.dev.git.n.h` format to be SemVer 2 compliant.
 1. If `--long` is specified or not. If `--long` is specified, tagged commits
    will always be written out with the `n.h` part appended to it, looking something
-   like `1.0.0-0git.0.habcd123`
+   like `1.0.0-0.dev.git.0.habcd123`
+
+When producing a development version (with `.git.n.hash` on the end),
+The _base_ version can come from one of two places,
+depending on your configuration.
+See [Controlling development version](#controlling-development-versions-in-chart.yaml) for more info.
 
 ### Examples chart versions and image tags
 
@@ -52,13 +57,13 @@ order that could come from using chartpress.
 
 ```
 0.8.0
-0.8.0-0git.4.hasdf123
-0.8.0-0git.10.hsdfg234
+0.8.0-0.dev.git.4.hasdf123
+0.8.0-0.dev.git.10.hsdfg234
 0.9.0-beta.1
-0.9.0-beta.1.0git.1.hdfgh345
-0.9.0-beta.1.0git.5.hfghj456
+0.9.0-beta.1.git.1.hdfgh345
+0.9.0-beta.1.git.5.hfghj456
 0.9.0-beta.2
-0.9.0-beta.2.0git.1.hghjk567
+0.9.0-beta.2.git.1.hghjk567
 0.9.0-beta.3
 0.9.0
 ```
@@ -170,6 +175,12 @@ charts:
     # --reset flag. It defaults to "0.0.1-set.by.chartpress". This is a valid
     # SemVer 2 version, which is required for a helm lint command to succeed.
     resetVersion: 1.2.3
+    # Use the version in Chart.yaml for the base version,
+    # instead of the latest tag from `git describe`
+    # This gives you more control over development version tags
+    # and lets you ensure prerelease tags are always sorted in the right order
+    useChartVersion: true
+
     # The git repo whose gh-pages contains the charts. This can be a local
     # path such as "." as well but if matching <organization>/<repo> will be
     # assumed to be a separate GitHub repository.
@@ -222,6 +233,83 @@ charts:
         skipPlatforms:
           - linux/arm64
 ```
+
+### Controlling development versions in Chart.yaml
+
+Like some "package version in version control" tools,
+you don't need to manage versions at all in chartpress except by tagging commits.
+
+However, relying only on tags results in "development versions" (versions published from commits after a release)
+having somewhat confusing prerelease versions.
+
+After publishing e.g. `1.2.3`, the next version will be `1.2.3-0.dev.git.1.habc`.
+According to Semantic Versioning,
+this is a "pre release" (good, it should be excluded from default installation),
+but it means it comes _before_ 1.2.3 (wrong! it's 1 commit _newer_ than 1.2.3).
+This is because prereleases should be defined relative to the _next_ release,
+not the _last_ release. But git tags only store the _last_ release!
+
+Chartpress 2.0 adds an option `chart.useChartVersion`,
+which changes the base version to use when setting the chart version.
+Instead of using the version of the last tag found via `git describe`,
+it uses the version found in Chart.yaml.
+
+The main benefits of this are:
+
+1. ensuring that published prerelease versions show up in the right order, and
+2. giving you control over the version of a prerelease chart (is it 2.0.0-0.dev or 1.3.1-0.dev?)
+
+Instead of publishing the sequence:
+
+- 1.2.3
+- 1.2.3-0.dev.git.1.habc (later than 1.2.3, but sorts _before_ 1.2.3!)
+
+You can publish
+
+- 1.2.3
+- 1.3.0-0.dev.git.1.habc (prerelease based on the _next_ version, not _last_)
+
+where chartpress will use the version in your Chart.yaml as the base version,
+instead of the last tag on the branch.
+
+This takes some extra configuration, and steps in your release process
+to update the version in your Chart.yaml.
+
+1. You must set `useChartVersion: true` in your chartpress.yaml:
+
+   ```yaml
+   charts:
+     - name: mychart
+       useChartVersion: true
+   ```
+
+2. You must control the version in Chart.yaml, especially after making a release.
+
+A release process would generally look like this:
+
+```bash
+V=1.2.3
+chartpress --reset --tag "$V"
+git commit -am "release $V"
+git tag -am "release $V" "$V"
+git push --atomic --follow-tags
+
+# back to development
+NEXT_V=1.3.0-0.dev
+chartpress --reset --tag "$NEXT_V"
+git commit -am "Back to $NEXT_V"
+git push --atomic
+```
+
+Any prerelease fields (such as `-0.dev` above, or `-alpha.1`)
+will be left as-is, with the `.git.n.hash` suffix added.
+
+If the version in Chart.yaml is not a prerelease,
+a prelease version will be constructed with `-0.dev.git.n.hash`.
+
+When you use this configuration,
+`chartpress --reset` strips the `.git...` suffix from the chart version,
+ignoring the `resetVersion` config.
 
 ## Caveats
 
--- a/README.md
+++ b/README.md
@@ -33,16 +33,13 @@ information.
       before its set as Helm 3 requires Helm chart versions to be SemVer2
       compliant.
 1. The latest commit modifying content in a _relevant path_ since `tag`.
-   1. `n`: The number of commits since the tagged commit, as an integer.
-   1. `h`: The latest commit's abbreviated hash. which is often 7-8 characters,
+   1. `n`: The number of commits since the latest tagged commit on the branch, as an integer.
+   1. `hash`: The latest commit's abbreviated hash, which is often 7-8 characters,
       prefixed with `h`.
-1. If `tag` (like `0.10.0` or `0.10.0-beta.1`) contains `-`,
-   indicating that it is a prerelease,
-   the chartpress suffix will be added to the prerelease fields,
-   e.g. `0.10.0-beta.1.git.5.habc123`
-   format will be used instead of a `tag-0.dev.git.n.h` format to be SemVer 2 compliant.
+1. If `tag` (like `0.10.0` or `0.10.0-beta.1`) contains a `-`, a `tag.git.n.hash`
+   format will be used, and otherwise a `tag-0.dev.git.n.hash` format will be used.
 1. If `--long` is specified or not. If `--long` is specified, tagged commits
-   will always be written out with the `n.h` part appended to it, looking something
+   will always be written out with the `.git.n.hash` part appended to it, looking something
    like `1.0.0-0.dev.git.0.habcd123`
 
 When producing a development version (with `.git.n.hash` on the end),
--- a/chartpress.py
+++ b/chartpress.py
@@ -787,8 +787,10 @@ def build_chart(
     Update Chart.yaml's version, using specified version or by constructing one.
 
     Chart versions are constructed using:
-        a) a base version, derived from either Chart.yaml or the latest tag
-        b) the latest commit that modified provided paths
+        a) a base version, derived from either Chart.yaml's version field or the latest git tag on branch
+        b) the latest commit that was tagged on the current branch (n)
+        c) the latest commit that modified provided paths (hash)
+        
 
     Example versions constructed:
         - 0.9.0-0.dev.git.2.hdfgh3456
@@ -1117,7 +1119,6 @@ def main(args=None):
     #   - push chart (--publish-chart, --extra-message)
     for chart in config["charts"]:
         forced_version = None
-        # TODO: maybe warn and switch default in the future?
         use_chart_version = chart.get("useChartVersion", False)
 
         if args.tag:
@@ -1126,7 +1127,7 @@ def main(args=None):
         elif args.reset and not use_chart_version:
             # resetting, get version from chartpress.yaml,
             # ignoring current version in Chart.yaml
-            forced_version = chart.get("resetVersion", "0.0.1-0.dev")
+            forced_version = chart.get("resetVersion", "0.0.1-set.by.chartpress")
 
         if not args.list_images:
             # update Chart.yaml with a version
--- a/chartpress.py
+++ b/chartpress.py
@@ -1076,7 +1076,7 @@ def main(args=None):
             # tag specified, use that version
             forced_version = args.tag
         elif args.reset and not use_chart_version:
-            # resetting, get version from chrtpress.yaml,
+            # resetting, get version from chartpress.yaml,
             # ignoring current version in Chart.yaml
             forced_version = chart.get("resetVersion", "0.0.1-0.dev")
 
--- a/chartpress.py
+++ b/chartpress.py
@@ -790,16 +790,23 @@ def build_chart(
             if "set.by.chartpress" in chart["version"]:
                 raise ValueError(
                     f"Chart {chart_file} has placeholder version {chart['version']}, with `useChartVersion: true`."
-                    f"Set the version in the {chart_file} to a base pre-release tag,"
-                    " e.g. `1.0.0-0.dev` or `1.0.0-alpha.1` and run chartpress again."
+                    f" Set the version in the {chart_file} to a base pre-release tag (your _next_ release),"
+                    " e.g. `2.0.0-0.dev` or `2.0.0-alpha.1` and run chartpress again."
                 )
 
             # require starting from a prerelease tag, to ensure correct ordering
             if "-" not in chart["version"]:
+                if reset:
+                    raise ValueError(
+                        "--reset after a release must also specify --tag $VERSION to reset to"
+                        " when using useChartVersion: true."
+                        " This should be your _next_ pre-release version,"
+                        " e.g. `chartpress --reset --tag 1.0.1-0.dev`"
+                    )
                 raise ValueError(
                     f"Chart {chart_file} has non-prerelease version {chart['version']} with `useChartVersion: true`"
-                    f"Set the version in the {chart_file} to a base pre-release tag,"
-                    " e.g. `1.0.0-0.dev` or `1.0.0-alpha.1` and run chartpress again."
+                    f" Set the version in the {chart_file} to a base pre-release tag (your _next_ release),"
+                    " e.g. `2.0.0-0.dev` or `2.0.0-alpha.1` and run chartpress again."
                 )
         else:
             base_version = None
--- a/chartpress.py
+++ b/chartpress.py
@@ -29,6 +29,9 @@ GITHUB_ACTOR_KEY = "GITHUB_ACTOR"
 # name of possible repository keys used in image value
 IMAGE_REPOSITORY_KEYS = {"name", "repository"}
 
+# prefix in prerelease tag for git commit count
+PRERELEASE_PREFIX = "0git."
+
 # Container builders
 class Builder(Enum):
     DOCKER_BUILD = "docker-build"
@@ -481,24 +484,28 @@ def _get_identifier_from_parts(tag, n_commits, commit, long):
     This function should provide valid Helm chart versions, which means they
     need to be valid SemVer 2 version strings. It also needs to return valid
     image tags, which means they need to not contain `+` signs either. We prefix
-    with n and h to ensure we don't get a numerical hash starting with 0 because
+    with h to ensure we don't get a numerical hash starting with 0 because
     those are invalid SemVer 2 versions.
+
+    We lead with '0git.' to ensure our suffixes sort as older than any explicit tags,
+    such as `0.1.2-alpha.1 > 0.1.2-alpha.0git.5.habc > 0.1.2-0git.3.h123`
 
     Example:
         tag="0.1.2", n_commits="5", commit="asdf1234", long=True,
-        should return "0.1.2-n005.hasdf1234".
+        should return "0.1.2-0git.5.hasdf1234".
     """
     n_commits = int(n_commits)
 
     if n_commits > 0 or long:
         if "-" in tag:
-            # append a pre-release tag, with a . separator
-            # 0.1.2-alpha.1 -> 0.1.2-alpha.1.n.h
-            return f"{tag}.n{n_commits:03d}.h{commit}"
+            # add a field to an existing prerelease tag
+            # 0.1.2-alpha.1 -> 0.1.2-{PRERELEASE_PREFIX}5.hsha
+            sep = "."
         else:
-            # append a release tag, with a - separator
-            # 0.1.2 -> 0.1.2-n.h
-            return f"{tag}-n{n_commits:03d}.h{commit}"
+            # create prerelease tag
+            # 0.1.2-alpha.1 -> 0.1.2-{PRERELEASE_PREFIX}5.hsha
+            sep = "-"
+        return f"{tag}{sep}{PRERELEASE_PREFIX}{n_commits}.h{commit}"
     else:
         return f"{tag}"
 
@@ -542,11 +549,11 @@ def build_images(
 
         Example 1:
         - long=False: 0.9.0
-        - long=True:  0.9.0-n000.hasdf1234
+        - long=True:  0.9.0-0git.0.hasdf1234
 
         Example 2:
-        - long=False: 0.9.0-n004.hsdfg2345
-        - long=True:  0.9.0-n004.hsdfg2345
+        - long=False: 0.9.0-0git.4.hsdfg2345
+        - long=True:  0.9.0-0git.4.hsdfg2345
 
     builder (str):
         The container build engine.
@@ -714,11 +721,11 @@ def build_chart(
 
     Example versions constructed:
         - 0.9.0-alpha.1
-        - 0.9.0-alpha.1.n000.hasdf1234 (--long)
-        - 0.9.0-alpha.1.n005.hsdfg2345
-        - 0.9.0-alpha.1.n005.hsdfg2345 (--long)
+        - 0.9.0-alpha.1.0git.0.hasdf1234 (--long)
+        - 0.9.0-alpha.1.0git.5.hsdfg2345
+        - 0.9.0-alpha.1.0git.5.hsdfg2345 (--long)
         - 0.9.0
-        - 0.9.0-n002.hdfgh3456
+        - 0.9.0-0git.2.hdfgh3456
     """
     # read Chart.yaml
     chart_file = os.path.join(name, "Chart.yaml")
--- a/chartpress.py
+++ b/chartpress.py
@@ -581,6 +581,9 @@ def build_images(
         The container build engine.
     platforms (list[str]):
         List of platforms to build for if not the same as the Docker host.
+    base_version (str):
+        The base version string (before '.git'), used when useChartVersion is True
+        instead of the tag found via `git describe`.
     """
     values_file_modifications = {}
     for name, options in images.items():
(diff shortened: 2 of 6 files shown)


minrk:
Actually, now that I think of it, our prefix _should_ come after alpha/beta/rc, because `1.2.3-$tag` means it's after `1.2.3` final, while `1.2.3-alpha.$tag` is before that (the fact that we can't use `+` means our post-release tags are always going to sort wrong relative to stable releases). This brings the same problem - `n` and `git` come after `beta`, but before `rc`. That's not the _most_ important thing, as we could recommend repos use `0dev` prerelease tags, which would solve the internal sorting as our tag fields would not be directly compared to alpha/beta/rc anymore.

minrk:
Another option is to use the Chart as the _source_ of the version. We are really publishing _pre_ release tags, whereas `git describe` produces _post_ release information. If we extracted the base version from Chart.version instead of `git describe`, we could:

1. avoid dev tags, which are an unusual practice
1. bump the base tag version in a PR (dev tags are always going to have a hard time because the tag must be coordinated with the merge of a 'bump-version' PR)
1. prerelease versions will be derived from the 'next' version, rather than the last version

consideRatio:
@minrk if we make a breaking change about this, I think I'd like us to make it change in a way that ensures that a commit beyond a tag leads to it being semver-sorted to also be a later release. So, not only ensure 1.2.3-n123 is sorted after alpha/beta/rc, but ensure its sorted after 1.2.3 - as a pre-release for 1.2.4.

`chartpress` makes assumptions about SemVer, because `helm` requires SemVer compliant versions. So, I think it could be reasonable to set versions like before where we automatically increment _the detected version_. Currently, we don't set the version field in Chart.yaml - instead we only set it when we publish the Chart based on detected tags on the current branch.

I still think there is some issue with our logic with detecting versions though whenever we work with backporting releases though and there is room for improvement about that.

## Brainstorming

__Loose thoughts:__

1. Chartpress ignores tags in other branches
2. We have released 1.2.0 of the z2jh helm chart, but 1.2.0 is a backport so it exists in `1.X`
3. The most common situation to push dev releases has been from the main branch - tagged releases have no version complexity.
4. I'd appreciate not having to update Chart.yaml versions with version controlled commits, but would accept them as a last resort fallback if needed to get robust logic.
5. Anything regarding this will be a breaking change that requires a new major release of chartpress

__Change idea:__

1. We let the automatic detected version be based on the semver latest git tag in the entire git repo.
2. We define new versions as a pre-alpha pre-release version based on an incremented patch version.
3. We still include the number of commits and the git hash, but we now need to handle counting a bit differently as the tag can be found in another branch that has diverged. Either we count the number of commits from the _base commit_ (first common commit they have) or add up the commits between both forked paths from the base commit.

minrk:
> So, I think it could be reasonable to set versions like before where we automatically increment the detected version

That's an interesting idea, though how do we decide what field to increment? E.g. in 1.x, it should be the 'minor' field, but in main it should be the major field. Maybe sometimes, it should be the patch field. Where should chartpress get that info? We could do major by default, and have a field in chartpress.yaml to specify differently?

Examples of (last tag) on (branch) and what the right chartpress tag should be (at least how we use it):

- 1.0 on main -> 2.0.0-n.h
- 1.0 on 1.x -> 1.1.0-n.h
- 2.0.0-alpha.1 on main -> 2.0.0-alpha.2.n.h (definitely not 3.0.0-n.h!)


> I still think there is some issue with our logic with detecting versions though whenever we work with backporting releases though and there is room for improvement about that.

I don't think there are any issues with our _detection_ of versions, but there is an issue in that the _most recent_ version is not in itself enough information to tell you what the next version is.

> We still include the number of commits and the git hash, but we now need to handle counting a bit differently as the tag can be found in another branch that has diverged

I don't think we should ever consider tags from another branch. I don't believe it provides any useful information, and traversing multiple git branches is getting a lot more complicated than I think we should consider.

A version like `2.0.0-5.habc` being 5 commits since 1.2.0 (or 1.0.0) ought to still be fine. All that matters is numbers go up. Adding a new tag on a backport 1.x branch shouldn't have any effect on the versions produced on the main branch.

consideRatio:
> [...] though how do we decide what field to increment? E.g. in 1.x, it should be the 'minor' field, but in main it should be the major field. Maybe sometimes, it should be the patch field. Where should chartpress get that info? We could do major by default, and have a field in chartpress.yaml to specify differently?

I'm thinking that this feature is specifically relevant for the default branch, not for the other branches. And, the default branch should be the latest. If that is the case, I would expect the most suitable version to mention would be the semver-latest git tag across the repo + incremented patch. If you are on a 1.X branch and chartpress is deciding on what to do etc, it could perhaps make sense to differ, but that doesn't happen often, but it would make sense to use the latest tag on that branch I guess.

I got to run - I'll followup!

consideRatio:
I'm leaning towards agreeing that using a tag from another branch is too problematic, and I'm opting away from thinking about that for now at least to reduce complexity.

We don't want version collisions, and if all calculated versions depend on the latest version on the branch, it means we should be fine unless we make the lowest increment we can.

__Minimally incremented version__

Most recent tag in branch | Minimally incremented version
-|-
`1.0.0` | `1.0.1`
`0.1.0` | `0.1.1`
`0.0.1` | `0.0.2`
`1.0.0-alpha.1` | `1.0.0-alpha.2`
`1.0.0-beta.1` | `1.0.0-beta.2`
`1.0.0-rc.1` | `1.0.0-rc.2`

I'm not sure though, what is semver-wise before and after with regards to `1.0.0-alpha.2` and `1.0.0-alpha.2.stuff`? It seems that appending `.stuff` made it a higher version. If our goal is to always increment versions with regards to semver, we could have an exception to minimally incremented versions for pre-release tags, and just append stuff to the tag directly.

image

minrk:
Thanks for thinking through it so thoroughly! I like that idea. I think the only thing not addressed by the minimal-increment is that it leaves us still needing to publish dev tags on z2jh if we want to signal when major changes happen in development, so that we can have 2.0.0-dev.n.h. I got excited that we might be able to solve that here, but it might be too much to ask for.

I agree that in practice, the relative ordering of git/alpha/beta doesn't come up often so we probably don't need to worry about it and can use the simplest "git".

consideRatio:
> I think the only thing not addressed by the minimal-increment is that it leaves us still needing to publish dev tags on z2jh if we want to signal when major changes happen in development, so that we can have 2.0.0-dev.n.h.

Hmmm, yes, we could have a strategy to push a pre-release tag to indicate a breaking change, but note that if you push `2.0.0-dev` you just pushed a tag that was sorted later than `2.0.0-alpha.1`. Even if chartpress would focus on the most recent tag, it would be causing weird behavior for any listing that sorted the tags based on semver version.

So in practice, if we want to signal breaking, and also use `2.0.0-alpha.1`, we would need something like `2.0.0-X` where `X is 0-9` or similar.

minrk:
WDYT about getting the version from Chart.yaml so that it's set by repo contents instead of `git describe`? We could still use git describe for the commit counter, but ignore the version it references, since the last version doesn't really tell you anything about the next version. "You tell us what the version should be"  seems simpler to get right than "We'll guess based on some heuristics". It fits into a pull-request workflow better than tags, which are pushed out-of-band. It does add a possible step to release, e.g. `chartpress --reset --tag 2.2.0-0.dev`.

manics:
https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/e7b93b4a6b110dd74bf4a8d00796cbe68ca61323/jupyterhub/Chart.yaml#L4 contains a version anyway
```yaml
version: 0.0.1-set.by.chartpress
```
So it's fine to include a placeholder tag already. Maybe we could keep that format, but make chartpress extract the base version (`0.0.1`)?

minrk:
> Maybe we could keep that format, but make chartpress extract the base version (0.0.1) in a well-defined manner?

Yeah, that's what I was thinking.

manics:
Another related idea- include something along the lines of [bump2version](https://pypi.org/project/bump2version/) or [tbump](https://pypi.org/project/tbump/) in chartpress to reduce the cognitive load in figuring out the version and ensuring you don't accidentally create a tag that doesn't appear in the correct. The above discussion about rc/alpha/beta etc is quite complicated, so being able to tell chartpress to bump a version based on a limited choice (e.g. major/minor/patch/dev/rc) could be helpful instead of making it completely free text as in https://github.com/jupyterhub/chartpress/pull/152?

minrk:
Yeah, integrating with tbump might be nice for setting the version.

consideRatio:
Hmmmm... It is a quite significant breaking change to require setting Chart.yaml version, and note that the dev versions is almost soleley relevant in the default branch that mat end up pushing dev releases.

What problem do we want to avoid? That the dev releases are versioned in a weird way?

Imagine we have a version from the chart, the version was 2.0.0-alpha.1 - a pre-release, should we append build suffix and create semver-later versions? If the version is 2.0.0, should we then append the build suffix and create semver-earlier versions, or make a minimal bump to 2.0.1 by chartpress?

If chartpress is to be version incremental free and append build suffixes systematically, it leads to that the UX must be that they should bump beyond current version sometimes but not all the times etc - depending on it being a pre-release or not.

Lets remember this change is only relevant for non-tagged commits, whenever there is a tagget commit, everything is clear - we use the version in the tag. So, this is solely a question about the version to use for dev releases.

manics:
We're going to have to define the dev version regardless of what versioning process we use, so I think defining it in `Chart.yaml` (and ideally building in a helper tool to bump it following whatever process we decide) is useful.

consideRatio:
I see myself pushing back on ideas at this point without really being sure i understand them.

I'd appreciate a very specific proposal on chartpress to consider more concretely. A proposal describing the logic on how: a) chartpress manipulates some version to set a dev version, b) chartpress decides the version to manipulate, and c) we should change release documentation in a helm chart repo using chartpress.

The "What i think currently" comment above still reflect my thinking.

minrk:
> Hmmmm... It is a quite significant breaking change to require setting Chart.yaml version,

It doesn't need to be breaking, if we make it opt-in, which is probably the safest approach.

>  and note that the dev versions is almost soleley relevant in the default branch that mat end up pushing dev releases.

Since publishing dev releases is the main reason chartpress was created (for z2jh, binderhub), I think it's appropriate to prioritize this use case, and [solve our own problems](https://github.com/jupyterhub/team-compass/issues/471).

> What problem do we want to avoid? That the dev releases are versioned in a weird way?

I think that's the main one, and there's _no input_ other than git when this makes a poor choice (as it does today for z2jh). The "right" place to put git-describe info in a version is in the "post-release" field, not pre, since it's counting after. Populating a 'prerelease' field needs to know what the next version is, and the previous version simply doesn't provide that information.


> If chartpress is to be version without version incremental logic and append build suffixes systematically to Chart.yaml specified version, it leads to that the UX becomes that the user should bump beyond current version sometimes but not all the times etc - depending on it being a pre-release or not - at least if we consistently want to release semver-newer versions.

It leads to the standard versioning workflow of releasing software packages, such as:

```
tbump 2.0.0 # set version, commit & tag
tbump --no-tag 2.1.0.dev # set dev version, commit
```

which might become

```
chartpress --tag 2.0.0 # set version
# git commit & tag
chartpress --reset --tag 2.1.0-0.dev # set dev version
# git commit
```

> I'd appreciate a very specific proposal on chartpress to consider more concretely.

Absolutely! I floated the idea to see if it made enough sense to pursue a more complete spec, but we should definitely do that.

minrk:
> A proposal describing the logic on how: a) chartpress manipulates some version to set a dev version

Given a base version (last tag from git describe, or 'current' version from chart):

1. if already a prerelease, append `.git.{n}.h{sha}` to the prerelease fields.
2. if not a prerelease, append `-0.dev.git.{n}.h{sha}` to ensure it sorts before '-alpha' (we could leave out '.dev')

Given 5 commits from the last tag and sha='abc', this would produce tags:

- 1.0.0 -> 1.0.0-0.dev.git.5.habc
- 1.0.0-xyz -> 1.0.0-xyz.git.5.habc
- 1.0.0-alpha.1 -> 1.0.0-alpha.1.git.5.habc
- 1.0.0-0.dev -> 1.0.0-0.dev.git.5.habc

Importantly, this _does not_ solve the "dev versions sort after releases" if you don't opt-in to the version-from-chart scheme _or_ use dev tags. This is the situation we have today, and I think it's appropriate to say that you should get the version from the chart if you want prereleases to be relative to the releases they are actually 'pre.' I don't think we should be guessing what the next release might be.

b) chartpress decides the version to manipulate,

opt-in, use the version in the chart exactly as-is (optionally: strip a standard '-set.by.chartpress' suffix):

```python
If `config["versionFromChart"]`:
    base_version = chart["version"]
else:
    base_version = last_tag_on_branch
```

> and c) we should change release documentation in a helm chart repo using chartpress.

Yes, I believe it should be:

```bash
V=2.0.0
# equivalent of tbump $V
chartpress --reset --tag "$V"
git commit -a -m "bump version to $V"
git tag -a -m "release $V" "$V"

NEXT_V = "2.1.0-0.dev"
# equivalent of tbump --no-tag $NEXT_V
chartpress --reset --tag "$NEXT_V"
git commit -a -m "bump version to $NEXT_V"

git push --atomic
# next published chart version will be 2.1.0-0.dev.git.1.habc
```

Optionally, we could have chartpress execute the git commit/tag, like tbump does.

You would get the _same versions_ if you skipped the versionFromChart scheme if you tagged the commit after release as `2.1.0-0.dev` (`n` would be lower by 1, but otherwise identical).

The NEXT_V step is only necessary for releases, not prereleases (alpha, beta, etc.). If you skip/omit the additional NEXT_V commit, the only downside is that your prereleases will evaluate below your last release (as is already the case today).

consideRatio:
> The "right" place to put git-describe info in a version is in the "post-release" field, not pre, since it's counting after. Populating a 'prerelease' field needs to know what the next version is, and the previous version simply doesn't provide that information.

I think post-release refers to `+` parts in a semver version, which is something `helm` won't handle well. For example, in a Helm chart repo index.yaml file, using `helm package` etc will update the index.yaml to remove the previous version I think. See discussion in https://github.com/helm/helm/issues/6710. With that considered, I think we must stick to using pre-release field for our dev versions as it is unacceptable to remove helm chart's we have published as we publish new ones.

### Helping chartpress with dev version information

I'm inclined to agree that yes - we must providing additional information to chartpress, and declaring a dev version somehow is the way to go about it.

1. We don't want to provide such information by creating git tags - that disrupts too much CI/CD automation and is a quite messy intervention.
2. Only the Helm chart's version needs to be updated, not the images.
   ```
   # this updates the version for Chary.yaml and values.yaml's image tags!
   chartpress --reset --tag 2.1.0-0.dev # set dev version
   ```
3. We have a pre-commit git hook in z2jh that makes us check that `chartpress --reset` doesn't lead to changes, so version controlling  `Chart.yaml` to some dev version will make us now need to adjust how we have used a pre-commit hook running `chartpress --reset`.

### New config in chartpress.yaml?

One idea i consider is to introduce new configuration in chartpress.yaml, that when set, will be used as a version to append build suffix to, whenever no explicit git tag is set.

Upsides to configuring version to use with build suffixes in chartpress.yaml?
- We avoid breaking current chartpress behavior
- `chartpress --reset` can remain used as before

On the other hand, perhaps directly setting a version in Chart.yaml would make sense? If we do that though:
1. `chartpress` must start respecting the configured Chart.yaml version at least some times and stop updating it in favor of the version itself concluded from git tag inspection. If chartpress is to respect the version in Chart.yaml, this can be done either a) always, or b) if chartpress.yaml config sais it should, or c) if Chart.yaml has a version value different from what `chartpress --reset` would reset it to, or d) some other condition.
2. `chartpress --reset` maybe shouldn't update the version entry in Chart.yaml as it did before depending on what we decide in the previous step.

minrk:
> I think post-release refers to `+` parts in a semver version, which is something `helm` won't handle well.

Yes, I should have been clearer. It's the right place semantically, but it's not available to us, which is why we have the issues we are having now - putting information relative to the last release in a field counting _down_ from it.

> Only the Helm chart's version needs to be updated, not the images.

True, and this is what `chartpress --reset --tag x` currently does (`chart.resetVersion` and `chart.resetTag` are set via different paths). This might be a bug! But we can make a new entrypoint, or add further clarification of what to do when, if that would be helpful.

> We have a pre-commit git hook in z2jh that makes us check that chartpress --reset doesn't lead to changes

I think `chartpress --reset` stripping the `.git.n...` suffix instead of totally resetting the whole version string ought to do the right thing here.

> One idea i consider is to introduce new configuration in chartpress.yaml, that when set, will be used as a version to append build suffix to, whenever no explicit git tag is set.

I think we do need something in chartpress.yaml to make the new behavior opt-in. I'm not sure it should be the version itself, though. I think that should be in the Chart.yaml, since there's already a field for it and we should use it. I think it should be your option b): "if chartpress.yaml config says it should" as that's the simplest, most explicit choice. Nothing to infer from something else.

> chartpress --reset maybe shouldn't update the version entry in Chart.yaml 

I think the right thing to do here is to strip the `.git...` suffix rather than clobber the whole string regardless of its value. This would also fix the `chartpress --reset` pre-commit issue.

README.md
```diff
@@ -170,6 +175,12 @@ charts:
     # --reset flag. It defaults to "0.0.1-set.by.chartpress". This is a valid
     # SemVer 2 version, which is required for a helm lint command to succeed.
     resetVersion: 1.2.3
+    # Use the version in Chart.yaml for the base version,
+    # instead of the latest tag from `git describe`
+    # This gives you more control over development version tags
+    # and lets you ensure prerlease tags are always sorted in the right order
```
consideRatio, line 181: Typo in prerelease here. I'm not sure why but I can't comment a suggestion.

EDIT: oops okay, because I'm reviewing the changes associated with a specific commit, not the full PR.

README.md
```diff
@@ -223,6 +234,83 @@ charts:
...
+
+Chartpress 2.0 adds an option `chart.useChartVersion`,
+which changes the base version to use when setting the chart version.
+Instead of using the version of the last tag found via `git describe`,
+it uses the version found in Chart.yaml.
+
+The main benefits of this are:
+
+1. ensuring that published prerelease versions show up in the right order, and
+2. giving you control over the version of a prerelease chart (is it 2.0-dev or 1.3-dev?)
```
consideRatio, line 260: Even if they are examples, let's avoid a non-semver version as they are strictly disallowed by `helm`. Adding a patch version identifier would solve it.

chartpress.py
```diff
@@ -1008,10 +1059,13 @@ def main(args=None):
...
     for chart in config["charts"]:
         forced_version = None
+        # TODO: maybe warn and switch default in the future?
+        use_chart_version = chart.get("useChartVersion", False)
+
         if args.tag:
             forced_version = args.tag
-        if args.reset:
-            forced_version = chart.get("resetVersion", "0.0.1-set.by.chartpress")
+        elif args.reset and not use_chart_version:
```
consideRatio, line 1067: Should this be `elif args.reset and use_chart_version` ?

tests/test_repo_interactions.py
```diff
@@ -30,12 +31,12 @@ def test_chartpress_run(git_repo, capfd):
...
     # summarize information from git_repo
     sha = git_repo.commit("HEAD").hexsha[:7]
-    tag = f"0.0.1-{PRERELEASE_PREFIX}1.h{sha}"
+    tag = f"0.0.1-{PRERELEASE_PREFIX}.1.h{sha}"
     check_version(tag)
 
     # run chartpress
     out = _capture_output([], capfd)
-
+    print(out)
```
consideRatio, line 39: Debugging statement remnant?

tests/test_repo_interactions.py
```diff
@@ -239,7 +237,7 @@ def test_chartpress_paths_configuration(git_repo, capfd):
     open("extra-chart-path.txt", "w").close()
     git_repo.git.add(all=True)
     sha = git_repo.index.commit("Added extra-chart-path.txt").hexsha[:7]
-    tag = f"0.0.1-{PRERELEASE_PREFIX}3.h{sha}"
+    tag = f"0.0.1-0.dev.git.3.h{sha}"
```
consideRatio, line 240: I see `{PRERELEASE_PREFIX}` be used sometimes, but not other times, that doesn't seem intentional to me - is it?

consideRatio (commented):
Not feeling confident about this still, but I made a review pass. By mistake I made comments associated with a specific commit instead of the entire PR =/

tests/test_repo_interactions.py
minrk: I kept changing my mind whether it was worth having as a constant, which is why I haven't been super consistent. I'm still not sure!

In the latest iteration, there are also two prefixes, so what should the variable be is ambiguous:

- `git.` - the prefix we _always_ use ahead of the `.{n}.h{sha}`
- `0.dev.git.` - the full prefix we use if there was no prerelease tag already

chartpress.py
minrk: No, this branch is "resetting chart version from chartpress.yaml, ignoring what's in Chart.yaml"

tests/test_repo_interactions.py
minrk: Yes and no - it's there for debugging, but I think it should stay. When this test fails, the output should be shown so you can see what's wrong (pytest capture means it won't be shown unless it fails).

tests/test_repo_interactions.py
minrk: Also ambiguous: should the constant include the trailing `.`? I think it reads better if it doesn't, but then can't be empty.

minrk: Went with `GIT_PREFIX="git"` and `PRERELEASE_PREFIX="0.dev.git"`, so excluding trailing `.`.

README.md
```diff
@@ -220,6 +234,83 @@ charts:
...
+
+However, relying only on tags results in "development versions" (versions published from commits after a release)
+having somewhat confusing prerelease versions.
+
+After publishing e.g. `1.2.3`, the next version will be `1.2.3-0.dev.git.1.habc`.
+According to Semantic Versioning,
+this is a "pre release" (good, it should be excluded from default installation),
+but it means it comes _before_ 1.2.3 (wrong! it's 1 commit _newer_ than 1.2.3).
+This is because prereleases should be defined relative to the _next_ release,
+not the _last_ release. But git tags only store the _last_ release!
```
consideRatio, line 250: This logic only holds for when using stable releases (`x.y.z` release, without pre-release sections), whenever we have `1.2.3-beta.1` for example, appending `.git.1.habcd123` makes it a more recent release.

The release process described below assumes you cut a stable release, but if you cut a pre-release like `1.2.3-beta.1` you shouldn't go back to `1.2.3-beta.2` or `1.2.3-beta.2` etc.

This complexity and UX is what makes me uneasy about accepting this change at this point. I perceive it as very complicated to document and convey the correct usage, to the extent I perceive it will be only you and me that understands how it works and how to use it reliably =/

README.md
```diff
@@ -220,6 +234,83 @@ charts:
...
+
+Chartpress 2.0 adds an option `chart.useChartVersion`,
```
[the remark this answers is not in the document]
minrk: Sorry, I don't know what you mean. What do you think the version should be?

chartpress.py
```diff
@@ -1001,10 +1069,16 @@ def main(args=None):
...
+
         if args.tag:
+            # tag specified, use that version
             forced_version = args.tag
-        elif args.reset:
-            forced_version = chart.get("resetVersion", "0.0.1-set.by.chartpress")
+        elif args.reset and not use_chart_version:
+            # resetting, get version from chrtpress.yaml,
+            # ignoring current version in Chart.yaml
+            forced_version = chart.get("resetVersion", "0.0.1-0.dev")
```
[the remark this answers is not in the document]
minrk: I don't think this is a breaking change at all. It does change what the version is reset _to_, but it's just as consistent. As you say, the version is never used, and after the chartpress version is upgraded once, it continues to be idemponent.

I can change it back, if you think `set.by.chartpress` is better.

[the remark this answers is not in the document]
consideRatio: Okay hmmm, yeah I think `set.by.chartpress` is better and I'm not sure what to be considered a breaking change - but I know this change would make the z2jh that relies on the default value need a change to help its CI system running `chartpress --reset` not report failures where `0.0.0-set.by.chartpress` as version controlled would change to `0.0.1-0.dev` instead.

I care mostly because I think the placeholder values in z2jh is more comprehensible like this than if they are set to `0.0.1-dev.0`, and, when this setting is disabled, they will be placeholders rather than a value to be used.

README.md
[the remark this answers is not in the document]
consideRatio: If this is a new feature without a breaking change that you opt-in to, then i figure we would make a minor version bump rather than a major version bump!

README.md
minrk: I get that. I want this to be understandable!

I know I'm more comfortable with it, because this is closer to how almost every _other_ project I work with is versioned - make a release, tag it, bump version to next-dev. You _may_ choose to skip the last step if you don't care about folks installing from a dev branch having sensible versions. I don't feel like I'm documenting special chartpress use, I'm documenting widespread standard practice, but the chartpress audience is a bit different.

I'm less concerned, too, because this is all opt-in. It's only important for folks who:

1. publish dev releases (us), and
2. care about dev releases having a sensible version (also us)

I think this is a better experience as both a maintainer and consumer of charts, and I think everyone _should_ opt-in to the new behavior, but the fact that it's optional does add to the complexity. I also want it to be clear that folks are welcome to ignore it - the only cost is improper ordering of prerelease versions (by a computer, humans can understand it fine!).

At a high level, I think the following workflow is not optional for folks who want to have development versions that make sense (whether a helm chart or any other package):

1. set release version
2. publish release (e.g. tbump, chartpress --tag)
3. (if stable release) set next-dev version

We have to consider exactly how to express that, but those three steps must exist somewhere, because there is no way to automate step 3 - the user has to tell _something_ what the next version is going to be. What do you think your ideal process would look like? Maybe we can make it work.

README.md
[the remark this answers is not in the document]
minrk: Got it, sorry. Since this does change the tag format (`-n005.habc` -> `-0.dev.git.5.habc`). I don't _think_ that'll matter to most folks, but it may enough that a major version feels appropriate.

chartpress.py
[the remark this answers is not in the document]
minrk: Fair enough, I'll put it back.

chartpress.py
```diff
@@ -727,7 +779,23 @@ def build_chart(
 
     # decide a version string
     if version is None:
-        version = _get_identifier_from_paths(*paths, long=long)
+        if use_chart_version:
+            # avoid adding .git... to chart version multiple times!
+            base_version = _trim_version_suffix(chart["version"])
+            # TODO: trim -set.by.chartpress?
+            # if base_version.endswith("-set.by.chartpress"):
+            #     base_version = base_version.rsplit("-", 1)[0]
```
consideRatio, line 787: My understanding is that this is a failsafe to avoid situations when Chart.yaml has been updated with a chartpress version suffix already.

In the past, running chartpress multiple times wouldn't cause longer and longer version strings as  the resulting Chart.yaml specified version wouldn't be used as input. But, it is now. Makes sense.

---

@minrk I think the suggestion below would make me more comfortable with the changes overall as I think it would reduce risk of incorrect use of chartpress and reduce complexity of chartpress code etc.

1. I suggest we don't trim any `-set.by.chartpress` stuff, but that we opt to error if we observer that the version string is matching `resetVersion`. That would indicate that use_chart_version is configured but a Chart.yaml file don't have a version specified yet.

2. I suggest that whenever use_chart_version is specified, we assert that Chart.yaml includes a pre-release version, and not a stable version.

   Consider the `_trim_version_suffix` function. Chartpress can appends `0.dev` automatically, but we don't trim that.
   
   Consider version sorting. If you have a stable version in Chart.yaml and haven't "returned to dev", then we end up with the same issue as before where version ordering would fail.

   These issues would go away if we don't let chartpress also append `0.dev` sometimes, but instead just require that when `use_chart_version` is set we also expect a pre-release to be specified in Chart.yaml. I think this would reduce complexity without restricting users, it just enforces the intended use I think.

minrk: I think I've implemented both of these with informative errors. I'm not sure that I agree that things are simplified by it, but there are fewer cases that are allowed to proceed without errors, at least.

minrk:
@consideRatio rebased after conflicts with #173

chartpress.py
```diff
@@ -1026,10 +1117,16 @@ def main(args=None):
     #   - push chart (--publish-chart, --extra-message)
     for chart in config["charts"]:
         forced_version = None
+        # TODO: maybe warn and switch default in the future?
```
consideRatio, line 1120: While we are not switching the default in this PR, should we plan for it?

- If chartpress is used to just publsih releases along with a git tag is pushed, then switching the default to `useChartVersion: true` is adding trouble.
- If using it to publish between git tags as well, then both options work, but I'm +1 for switching the default.

I understand it as switching the default would imply forcing simple use cases (jupyterhub/pebble-helm-chart, dask/dask-gateway, and dask/dask-kubernetes) would need to manually declare the Chart.yaml version or learn about this and opt out even though they don't need the functionality.

I'm inclined to think we shouldn't switch, but instead ask that those that wants to emit dev releases in this more nice way also take on the work of maintaining Chart.yaml's version themselves.

```suggestion
```

chartpress.py
```diff
@@ -1026,10 +1117,16 @@ def main(args=None):
...
+
         if args.tag:
+            # tag specified, use that version
             forced_version = args.tag
-        elif args.reset:
-            forced_version = chart.get("resetVersion", "0.0.1-set.by.chartpress")
+        elif args.reset and not use_chart_version:
+            # resetting, get version from chartpress.yaml,
+            # ignoring current version in Chart.yaml
+            forced_version = chart.get("resetVersion", "0.0.1-0.dev")
```
consideRatio, line 1129: This would be another small breaking change, so I'd prefer to let this remain as it was to not complicate things.

```suggestion
            forced_version = chart.get("resetVersion", "0.0.1-set.by.chartpress")
```

consideRatio (commented):
Thanks @minrk for your work on this!

I think this is acceptable pretty much as it is, I left some smaller comments and I didn't review the tests but I hope its okay to go for anyhow.

chartpress.py
```diff
@@ -725,23 +759,44 @@ def _update_values_file_with_modifications(name, modifications):
...
+    strict_version=False,
+    use_chart_version=False,
 ):
     """
     Update Chart.yaml's version, using specified version or by constructing one.
 
     Chart versions are constructed using:
-        a) the latest commit that is tagged,
+        a) a base version, derived from either Chart.yaml or the latest tag
         b) the latest commit that modified provided paths
```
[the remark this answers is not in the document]
consideRatio:
```suggestion
        a) a base version, derived from either Chart.yaml's version field or the latest git tag on branch
        b) the latest commit that was tagged on the current branch (n)
        c) the latest commit that modified provided paths (hash)
        
```

chartpress.py
minrk: Let's make it purely opt-in for now, test it out a few places (e.g. z2jh), and see how we feel. If it seems like a pure win, we can make a patch release notifying that we'll change the default, otherwise let's leave it alone.

chartpress.py
```diff
@@ -750,7 +807,43 @@ def build_chart(
...
+        else:
+            base_version = None
+        if reset:
+            # resetting, use the base version without '.git...'
+            version = base_version
+        else:
+            # derive the full version, with possible '.git...'
+            version = _get_identifier_from_paths(
+                *paths, long=long, base_version=base_version
+            )
```
consideRatio, line 845: Hmmmm the `if reset:` clause above is evaluated only when `version is None`. But it seems that if `reset` is True, then `build_chart()` is always called with the `version` parameter set based on this, where we see the only call to `build_chart`.

```python
        if args.tag:
            # tag specified, use that version
            forced_version = args.tag
        elif args.reset and not use_chart_version:
            # resetting, get version from chartpress.yaml,
            # ignoring current version in Chart.yaml
            forced_version = chart.get("resetVersion", "0.0.1")

        if not args.list_images:
            # update Chart.yaml with a version
            chart_version = build_chart(
                chart["name"],
                paths=_get_all_chart_paths(chart),
                version=forced_version,
                long=args.long,
                reset=args.reset,
                strict_version=args.publish_chart,
                use_chart_version=use_chart_version,
            )
```

Leaving this in, even though its currently unreachable is also not great because if it ever becomes reachable, then version could end up as `None` after this code segment, and then finally the Chart.yaml etc could get a version field set to `None` as well.

consideRatio (dismissed):
I found logic introuced that with the `chartpress` CLI is unreachable that I'm not sure how it makes sense to rework.

chartpress.py
consideRatio: I think the following change is equivalent to what we have now, because of the unreachability of the `if` section.

```suggestion
        # derive the full version, with possible '.git...'
        version = _get_identifier_from_paths(
            *paths, long=long, base_version=base_version
        )
```

minrk: This isn't unreachable, it's the path taken by `chartpress --reset` (e.g. as called by the pre-commit hook) when use_chart_version is True. I'll try to clear up the multiple paths, probably by moving the selection of the reset version nearby so they two paths aren't so different.

minrk:
@consideRatio I tried to make the reset logic clearer by doing more in `build_chart` so that the version to reset to is set in the same place, whether useChartVersion is true or false.

I also added more test coverage, including for #174 to show that it is indeed fixed.

chartpress.py
consideRatio: I didn't understand that you are right about this, but reviewing the new changes makes me fine about the current state which seems good to me - thank you!

consideRatio (approved):
Thank you @minrk!!!

minrk:
Thanks for your thoughtful review!

consideRatio:
For visibility, I've opened #176 and #177 as followups to this PR.

PR #152, opened 2022-03-02
accept --tag arg in --reset

minrk:
allows `chartpress --reset --tag 2.0.0-dev`. Related to #150


consideRatio (approved):
Fine frature that LGTM no matter what chartpress UX we settle for.

PR #162, opened 2022-06-10
Only push published chart if --push specified

manics:
It's useful when developing a new chart to be able to test the publish chart functionality without pushing the `gh-pages` branch.

I'm using the existing `--push` argument since it doesn't seem like there's a use for pushing the chart unless the images are already pushed, though it could be a separate argument instead?

### DIFF
--- a/chartpress.py
+++ b/chartpress.py
@@ -549,7 +549,7 @@ def build_images(
         If unspecified the tag for each image will be the hash of the last commit
         to modify the image's files.
     push (bool):
-        Whether to push the resulting images (default: False).
+        Whether to push the resulting images and chart if published (default: False).
     force_push (bool):
         Whether to push the built images even if they already exist in the image
         registry (default: False).
@@ -773,6 +773,7 @@ def publish_pages(
     chart_repo_url,
     extra_message="",
     force=False,
+    push=True,
 ):
     """
     Update a Helm chart registry hosted in the gh-pages branch of a GitHub git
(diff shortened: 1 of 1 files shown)


consideRatio:
Hmmm...

We have the following CLI flags.

```
  --push                Push built images to their image registries, but not
                        if it would replace an existing image.
  --force-push          Push built images to their image registries,
                        regardless if it would replace an existing image.
  --publish-chart       Package a Helm chart and publish it to a Helm chart
                        registry constructed with a GitHub git repository and
                        GitHub pages, but not if it would replace an existing
                        chart version.
  --force-publish-chart
                        Package a Helm chart and publish it to a Helm chart
                        registry constructed with a GitHub git repository and
                        GitHub pages, regardless if it would replace an
                        existing chart version
```

What are you suggesting?

manics:
> What are you suggesting? Are you looking for an option to package but not publish the chart?

Yes! It's useful during development of a new chart, and also when developing new chartpress features.

consideRatio:
I'm hesitant about this PR as it is, thinking that the implication of what the `--push` flag is stretched too far to be comprehensible for users and maintainers. While the name `--push` is general, it means to me `--push-images` and should be treated as that.

> I'm using the existing --push argument since it doesn't seem like there's a use for pushing the chart unless the images are already pushed, though it could be a separate argument instead?

Thanks for highlighting this assumption! You can have a chart that doesn't include images to be built. Due to that, this is a breaking change because it can make sense for that chart to do `--publish-chart` without `--push`.

I think having a `--package-chart` flag, which is implied by `--publish-chart`/`--force-publish-chart`, would be the way to go for this.

consideRatio:
I propose we implement this via `--package-chart` instead, what do you think @manics?

```
--package-chart
--publish-chart        # implies --package-chart
--force-publish-chart  # implies --publish-chart
```

Issue #177, opened 2022-08-28
Issues remain for handling development releases


consideRatio:
### Background

`chartpress` has relied on information in git tags to deduce development chart versions and image tags, but we realized a need to guide chartpress more explicitly and provide it a _base version_ of some kind. In #150 we introduced the `useChartVersion` option that made Chart.yaml's version field be the source of the _base version_.

Keeping state in Chart.yaml's version field is a bit complicated though.
- Having Chart.yaml's version field as the source of the base version, but not exactly the base version, makes us need to handle logic to deduce the base version.
- With #150 we made `chartpress --reset` more complicated and influenced by useChartVersion` being true or false.
- With #150 we end with a workflow that pushes default versions and values that are not published. For example, we would have `2.0.0-0.dev` as a version in z2jh's Chart.yaml's version field (and image tags), but they are not available. It could be less confusing if it instead said something like `set-by-chartpress` to indicate a need to use the tool chartpress.

### Alternative proposal (1)

- We start speaking about a _base version_, which typically chartpress has calculated as the latest git tag on branch or `0.0.1` as a fallback.
  - Benefit: easy to understand this concept
- We let the _base version_ be possible to explicitly override in `chartpress.yaml` under `<chart>.baseVersion`.
  - Benefit: I think it will be relatively easy to understand this configurations impact, as well as implementing it and maintaining it
- We require `<chart>.baseVersion` to be a pre-release if specified, and document examples like `2.0.0-0.dev` or `2.0.0-beta.1` 
  - Benefit: we end up not needing logic to add `-0.dev` in chartpress itself
  - Benefit: we end up being able to speak about a _chartpress suffix_ or _chartpress development release suffix_ to mean specifically the `.git.n.hash` part and not add confusion about it also meaning `-0.dev` or similar.

#### Proposed workflow with development releases

- Initially set `baseVersion` in `chartpress.yaml` to your planned next release.

  ```yaml
  charts:
    - name: mychart
      baseVersion: 1.2.0-0.dev
  ```

- When a release is to be made associated with a git tag, just push a git tag and trigger automation configured to run `chartpress --push --publish-chart --tag <tag>`

  ```bash
  V=1.2.0
  git commit -am "release $V"
  git tag -am "release $V" "$V"
  git push --atomic --follow-tags
  ```

- After a major or minor release (such as 1.2.0) or because of a change of plans, update the configured base version to reflect the next minor version you plan to release (such as 1.3.0, or 2.0.0).

[...]

PR #178, opened 2022-08-29
don't reset commit count on each tag

minrk:
always count from the beginning, ensuring the count increases no matter what the base version or recent tag.

I believe this also resolves the 'implicit tag' issue in #177. We already had a [test](https://github.com/jupyterhub/chartpress/blob/main/tests/test_repo_interactions.py#L110) for this, but it didn't ensure the repo was clean, so useChartVersion passed (and it would have worked, too, since the useChartVersion workflow means committing changes to the Chart.yaml).

I'll open a separate PR trading useChartVersion for baseVersion in chartpress.yaml

### DIFF
--- a/chartpress.py
+++ b/chartpress.py
@@ -511,6 +511,7 @@ def _get_identifier_from_paths(*paths, long=False, base_version=None):
         if base_version is None:
             base_version = latest_tag_in_branch
     except subprocess.CalledProcessError:
+        # no tags on branch
         pass
 
     if base_version is None:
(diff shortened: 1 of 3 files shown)


chartpress.py
```diff
@@ -484,30 +495,28 @@ def _get_identifier_from_paths(*paths, long=False, base_version=None):
...
-        n_commits = int(
-            _check_output(
-                ["git", "rev-list", "--count", latest_commit],
-                echo=False,
-            )
-            .decode("utf-8")
-            .strip()
-        )
-        sha = latest_commit
+        pass
```
consideRatio, line 514:
```suggestion
        # no tags on branch
        pass
```

consideRatio (approved):
Added back parts of a comment, but this LGTM.

What do you think about renaming the title of this PR to focus on what now is done? I hope that this title suggestion helps readers of the changelog understand the change better.

From "don't reset commit count on each tag" to "count commits on branch instead of commits since last tag on branch"

minrk:
Since the numbers aren't functional until the next release (that's what #150 fixes), I'd call it maintenance/enhancement.

PR #180, opened 2022-08-30
Enforce that `--reset` is an exclusive argument

### LINKED ISSUE #176: `--reset --tag` takes on the existing logic of `--skip-build --tag`
I think the combination of #152 and #150 made `chartpress --reset` more complicated than it should be, because the functionality we ended up with is the same as `chartpress --skip-build` when both are given `--tag`.

Practically, I'm thinking that the commands below are the same.

```
chartpress --reset --tag 1.2.3
chartpress --skip-build --tag 1.2.3
```

Should we revert the logic of `--reset --tag` as implemented via #152 and #150?

Ping to @minrk for visibility, and also #174 is worth a mention from this issue.

minrk:
avoids confusion with `--reset --tag` which should be `--no-build --tag`

This could be done more robustly by switching to subcommands, but I'm not sure now is the time to make that change.

closes #176

### DIFF
--- a/chartpress.py
+++ b/chartpress.py
@@ -8,6 +8,7 @@ import argparse
 import os
 import pipes
 import re
+import shlex
 import shutil
 import subprocess
 import sys
@@ -1038,7 +1039,7 @@ class ActionAppendDeprecated(argparse.Action):
         getattr(namespace, self.dest).append(values)
 
 
-def main(args=None):
+def main(argv=None):
     """Run chartpress"""
     argparser = argparse.ArgumentParser(description=__doc__)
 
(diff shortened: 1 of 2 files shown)


chartpress.py
```diff
@@ -1127,24 +1128,33 @@ def main(args=None):
     )
     argparser.add_argument(
         "--version",
-        action="store_true",
-        help="Print current chartpress version and exit.",
+        action="version",
```
minrk, line 1131: argparse already has a 'version' action that does what we want.

consideRatio (approved):

consideRatio:
Thank you @minrk!!!

PR #190, opened 2022-10-27
Add BRANCH expansion

bleggett:
Much like `TAG` or `LAST_COMMIT`, add an expansion for `BRANCH` to expand to the current Git branch.

This is useful if you are using OCI image caching via extraBuildArgs, like so:
```
        extraBuildCommandOptions:
          - --cache-to
          - type=registry,ref=ghcr.io/myorg/myimg-buildcache:{BRANCH}
          - --cache-from
          - type=registry,ref=ghcr.io/myorg/myimg-buildcache:{BRANCH}
```

PR branches won't have TAG for us, and LAST_COMMIT will create new caches for every new Git SHA, which defeats the purpose of a `docker buildx` cache save/restore.

NOTE - I cannot get the existing tests on `main` to pass locally with current Docker release, as the default `docker build` output format doesn't match what the tests expect - if I am missing something here let me know, otherwise I will try to debug any test failures via CI.

### DIFF
--- a/chartpress.py
+++ b/chartpress.py
@@ -275,7 +275,6 @@ def _get_current_branchname(**kwargs):
     )
 
 
-
 def _get_image_build_args(image_options, ns):
     """
     Render buildArgs from chartpress.yaml that could be templates, using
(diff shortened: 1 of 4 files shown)


minrk:
Thanks!

re: local build failures - yes, that's the changed output in buildkit. We need to fix that (#191), but as long as CI passes, we're okay for now. I beleive if you set `DOCKER_BUILDKIT=0` you'll get tests passing locally.

PR #229, opened 2024-01-06
`chartPath`: Chart directory can be different from the chart name

manics:
This makes it easier to organise charts in a git repo, for example the chart could be under `PROJECT-NAME/helm-chart` instead of `PROJECT-NAME/helm-chart/PROJECT-NAME`, whilst still keeping chartpress.yaml at the root of the repository.

This is a Helm 3 feature https://github.com/helm/helm/issues/1979 (should we drop Helm 2 instead of having a conditional test?)

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -101,6 +101,8 @@ jobs:
       - name: Run tests
         run: |
           pytest --verbose --color=yes --cov chartpress
+        env:
+          HELM2: ${{ matrix.helm2 }}
 
       # GitHub action reference: https://github.com/codecov/codecov-action
       - uses: codecov/codecov-action@v3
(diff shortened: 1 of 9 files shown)


consideRatio (approved):
Nice, this looks great - with the following details, this is approved!

- [x] Title to mention `chartPath`
- [x] Readme's configuration header to mention chartPath

---

> This is a Helm 3 feature https://github.com/helm/helm/issues/1979 (should we drop Helm 2 instead of having a conditional test?)

I think this is fine, but its also fine to drop support for helm2 - its ancient atm.

manics:
Done! Thanks for reviewing.

consideRatio (approved):
Beautiful PRs @manics!

PR #230, opened 2024-01-06
Autoincrement base version from tag if not specified

manics:
Currently Chartpress defaults to using the last existing tag with a dev suffix, but with semver this means it's older than the last tag.

This new option means the base version can be automatically inferred by incrementing the specified component (major/minor/patch) of the last tag, meaning it's no longer necessary to set base-version to the next tag.

https://github.com/yuvipanda/cryptnono/pull/12

The default behaviour is unchanged, though arguably we could default to `patch`?

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,8 +182,10 @@ charts:
     # if baseVersion is not a prerelease version (no -suffix),
     # the suffix `-0.dev` will be appended.
     #
-    # Alternatively this can be "major", "minor", or "patch", which will set
-    # baseVersion by incrementing the latest tag in the specified component.
+    # Alternatively baseVersion can be set to "major", "minor", or "patch", then
+    # baseVersion will be calculated based on the latest version tag from `git
+    # describe`, but have its "major", "minor", or "patch" version increment if
+    # the version isn't a pre-release.
     baseVersion: 3.2.1-0.dev
 
     # The git repo whose gh-pages contains the charts. This can be a local
--- a/README.md
+++ b/README.md
@@ -181,6 +181,9 @@ charts:
     # Recommended together with a version-bumping tool like `tbump`.
     # if baseVersion is not a prerelease version (no -suffix),
     # the suffix `-0.dev` will be appended.
+    #
+    # Alternatively this can be "major", "minor", or "patch", which will set
+    # baseVersion by incrementing the latest tag in the specified component.
     baseVersion: 3.2.1-0.dev
 
     # The git repo whose gh-pages contains the charts. This can be a local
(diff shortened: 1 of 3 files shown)


minrk (approved):

tests/test_helpers.py
```diff
@@ -183,17 +183,28 @@ def test__get_image_extra_build_command_options(git_repo):
         ("1.2.3-0.dev", "2.0.0", 0, "1.2.3-0.dev"),
         # ignore invalid semver tags
         ("1.2.3-0.dev", "x.y.z", 10, "1.2.3-0.dev"),
+        # autoincrement latest tag
+        ("major", "1.2.3", 10, "2.0.0-0.dev"),
+        ("minor", "1.2.3", 10, "1.3.0-0.dev"),
+        ("patch", "1.2.3", 10, "1.2.4-0.dev"),
```
consideRatio, line 189: What is the intended behavior if the latest tag is a pre-release like `1.2.3-beta.1`?

I'm thinking the point of autoincrementing is to avoid the incorrect ordering by appending `-0.dev`, but appending more characters to a pre-release like `1.2.3-beta.1` will not lead to incorrect ordering. So, we don't auto-increment the latest tag if its `1.2.3-beta.1` right?

README.md
```diff
@@ -181,6 +181,9 @@ charts:
     # Recommended together with a version-bumping tool like `tbump`.
     # if baseVersion is not a prerelease version (no -suffix),
     # the suffix `-0.dev` will be appended.
+    #
+    # Alternatively this can be "major", "minor", or "patch", which will set
+    # baseVersion by incrementing the latest tag in the specified component.
```
consideRatio, line 186: This didn't read so clearly to me yet, here is what I think could be a refinement.

I made an attempt below.

```suggestion
    # Alternatively baseVersion can be set to "major", "minor", or "patch", then
    # baseVersion will be calculated based on the latest version tag from `git
    # describe`, but have its "major", "minor", or "patch" version increment if
    # the version isn't a pre-release.
```

tests/test_helpers.py
consideRatio: Looking at the current implementation, I understand this test case to pass:

```
("patch", "1.2.3-beta.1", 10, "1.2.4-0.dev"), # but I expected `1.2.3-beta.1` here
```

consideRatio (commented):
I think this makes

manics:
Thanks for reviewing! I've updated this with your feedback.

consideRatio (approved):
Nice!

PR #234, opened 2024-03-26
Allow specifying config file with cli option (#170)

### LINKED ISSUE #170: add a `--file` flag to use a different `chartpress.yaml`
### Proposed change
In projects involving more than one chart, it would be useful to have a `--file` flag such that a specific `chartpress.yaml` file could be targeted, e.g. 

```
chartpress --file another-chartpress.yaml
```
### Alternative options
This behavior can be achieved by separating out charts via the directory structure and running `chartpress` in subdirectories, but this is sub-optimal, tedious and error-prone. 

### Who would use this feature?
Anyone building multiple charts in a single repo that don't all need to be built/packaged at the same cadence.

adamblake:
This adds a config option `--config` which accepts a file path to an alternative configuration file (e.g. `chartpress.dev.yaml`).  The PR passes all existing tests and adds a few new assertions to test the new code.

The use-case I am supporting is alternative chartpress files for different deployment contexts. For example, I have a local development context for a JupyterHub that has a couple extra convenience services, and those need different build arguments for the Hub Docker image. The way that I manage this currently is to temporarily rewrite `chartpress.yaml` with the updated values before calling it, but I would instead like to just load a different configuration file.

(The ideal implementation would be something more akin to how you can layer different configs on like you can with Helm values, but that's something I don't really have time to develop.)

After checking the issue lists, I believe this would also close #170

### DIFF
--- a/chartpress.py
+++ b/chartpress.py
@@ -4,6 +4,7 @@ Automate building and publishing helm charts and associated images.
 
 This is used as part of the JupyterHub and Binder projects.
 """
+
 import argparse
 import os
 import pipes
@@ -277,13 +278,13 @@ def _get_current_branchname(**kwargs):
 
 def _get_image_build_args(image_options, ns):
     """
-    Render buildArgs from chartpress.yaml that could be templates, using
+    Render buildArgs from the config file that could be templates, using
     provided namespace that contains keys with dynamic values such as
     LAST_COMMIT or TAG.
 
     Args:
     image_options (dict):
-        The dictionary for a given image from chartpress.yaml.
+        The dictionary for a given image from the config file.
         Fields in `image_options['buildArgs']` will be rendered
         and returned, if defined.
     ns (dict): the namespace used when rendering templated arguments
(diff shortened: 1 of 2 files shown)


chartpress.py
```diff
@@ -1307,5 +1324,19 @@ def main(argv=None):
             )
 
 
+def remove_config_arg(argv):
```
manics, line 1327: Can you make this return a copy instead of modifying the list in place? This will be easier to reason about/maintain in future. I think it'd be also worth renaming it `_remove_config_arg` since there's no reason to want to call this from outside.

tests/test_repo_interactions.py
```diff
@@ -80,6 +91,23 @@ def test_chartpress_run(git_repo, capfd, base_version):
         f"Updating testchart/values.yaml: image: testchart/testimage:{reset_tag}" in out
     )
 
+    # verify usage of --config
+    with temporary_copy("chartpress.yaml") as temp:
```
manics, line 95: It looks like this copy isn't used anywhere else, so can it be a separate testcase? Or parameterise the whole test (so the whole of test_chartpress_run is run with chartpress.yaml and a custom-named config file)?

tests/test_repo_interactions.py
```diff
@@ -9,6 +11,15 @@
 from chartpress import PRERELEASE_PREFIX, yaml
 
 
+@contextlib.contextmanager
```
manics, line 14: We've got several pytest fixtures in https://github.com/jupyterhub/chartpress/blob/main/tests/conftest.py which behave as context managers, would you mind following that pattern?

manics (commented):
Thanks, this looks like a nice addition

tests/test_repo_interactions.py
adamblake: Great idea to parameterize

tests/test_repo_interactions.py
adamblake: Sure thing

adamblake: Parameterizing removed the need for this, so I just removed it

adamblake:
@manics this is now refactored per your requests -- thanks!

manics (approved):
Thanks, sorry for the delay

PR #247, opened 2025-01-21
Transition to use pyproject.toml and tbump

consideRatio:
With this, we move away from setup.py and bump2version.


minrk:
Thanks!

